PR #4, opened 2017-06-15
add for rest api to get last certificate for given userid

lifemagazine:
In some cases, fabric members need to get the certificate of other members, not ca certificate.
So I appended the rest api to retrieve the certificate by user id.


lifemagazine:
In some cases, fabric members need to get the certificate of other members, not ca certificate.
So I appended the rest api to retrieve the certificate by user id.

PR #7, opened 2018-03-24
Update dbutil.go

lihaoguo:
Modify the field name length of the table affiliations to fix the following bug:
Error occurred initializing database: Failed to create user registry for MySQL: Failed to create MySQL tables: Error creating index on affiliations table: Error 1071: Specified key was too long; max key length is 767 bytes


ryjones:
Thanks for the contribution! this is a read-only mirror. please read the [contribution guide](http://hyperledger-fabric.readthedocs.io/en/release-1.1/CONTRIBUTING.html).

PR #13, opened 2019-08-30
fix the Ca port problem in the Enroll Orderer chapter

xiaomdong:
Org's CA server port is 7053
      [INFO] Listening on https://0.0.0.0:7053
But  in Enroll Orderer chapter , the Org's CA server port is 7056 
      fabric-ca-client enroll -d -u https://orderer-org0:ordererPW@0.0.0.0:7056

so this is a bug


xiaomdong:
again

PR #15, opened 2019-09-07
fiexd mistakes added config.yml  git commit --amend --signoff Sajan Jaiswal

idiota101:
Signed-off-by: Sajan Jaiswal sjnjaiswal@gmail.com
while following through the CA operations guide I found some typing mistakes and also a missing configuration i.e config.yml in every organisation MSP, I also found that the admin identity is beings registered with "id.type user " which is not correct, so I  have changed that as well.


idiota101:
git commit --amend --signoff sajan Jaiswal

PR #18, opened 2019-09-16
[FABCI-420] Cleanup AZP definition file in release-1.4

btl5037:

### DIFF
--- a/ci/azp-pipeline.yml
+++ b/ci/azp-pipeline.yml
@@ -14,8 +14,8 @@ pr:
 
 variables:
   GOPATH: $(Agent.BuildDirectory)/go
-  PATH: /bin:/usr/bin:/sbin:/usr/sbin:/usr/local/bin:/usr/local/sbin
-  GOVER: 1.12.5
+  PATH: $(Agent.BuildDirectory)/go/bin:/bin:/usr/bin:/sbin:/usr/sbin:/usr/local/bin:/usr/local/sbin
+  GOVER: 1.11.5
 
 jobs:
 - job: VerifyBuild
@@ -25,6 +25,8 @@ jobs:
   timeoutInMinutes: 60
   steps:
   - template: install_deps.yml
+  - script: go get golang.org/x/tools/cmd/goimports
+    displayName: Download goimports
   - checkout: self
     path: 'go/src/github.com/hyperledger/fabric-ca'
     displayName: Checkout Fabric CA Code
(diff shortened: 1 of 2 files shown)


ci/azp-pipeline.yml
```diff
@@ -25,6 +25,8 @@ jobs:
   timeoutInMinutes: 60
   steps:
   - template: install_deps.yml
+  - script: go get golang.org/x/tools/cmd/goimports
```
sykesm: This is pulling from master, something that has been the source of many issues as the formatting rules change fairly regularly.  To avoid problems, we vendor a version[1] in fabric and install it from the tree[2].

[1]: https://github.com/hyperledger/fabric/blob/0833f035234ff881756a179bddba981dcbe398d2/Gopkg.toml#L7
[2]: https://github.com/hyperledger/fabric/blob/0833f035234ff881756a179bddba981dcbe398d2/gotools.mk#L38-L41

ci/azp-pipeline.yml
```diff
@@ -53,10 +55,12 @@ jobs:
...
+    displayName: Install Tox
   - checkout: self
     path: 'go/src/github.com/hyperledger/fabric-ca'
     displayName: Checkout Fabric CA Code
   - script: tox -edocs
     displayName: Build Documentation
-  - script: '$(GOPATH)/go/src/github.com/hyperledger/fabric-ca/docs/_build/html'
-    displayName: Publish Documentation
\ No newline at end of file
+  - publish: '$(GOPATH)/src/github.com/hyperledger/fabric-ca/docs/_build/html'
```
sykesm: Why are we referencing GOPATH? Why isn't this relative to the build directory or working directory?

ci/azp-pipeline.yml
btl5037: Will address in a separate PR

ci/azp-pipeline.yml
btl5037: Changed to relative to working directory

btl5037:
> This is a reasonable start. Please squash the commits and remove the reference to GOPATH. We can address goimports (and other tools) in a separate PR.
> 
> Thanks.

Addressed

ci/azp-pipeline.yml
```diff
@@ -14,8 +14,8 @@ pr:
 
 variables:
   GOPATH: $(Agent.BuildDirectory)/go
-  PATH: /bin:/usr/bin:/sbin:/usr/sbin:/usr/local/bin:/usr/local/sbin
-  GOVER: 1.12.5
+  PATH: $(Agent.BuildDirectory)/go/bin:/bin:/usr/bin:/sbin:/usr/sbin:/usr/local/bin:/usr/local/sbin
+  GOVER: 1.11.5
```
sykesm: I don't remember this change last time. Looks like ci.properties was coded to use 1.12.5.

PR #19, opened 2019-09-16
[FABCI-420] Cleanup AZP definition file in master

btl5037:


btl5037:
> Looks like this is the same as #18 so the comments there apply here as well.

Addressed per #18 comments

PR #25, opened 2019-09-24
[FABCI-419] Version Go Tools

btl5037:
Use the tools.go model to version tooling deps

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,6 +28,7 @@
 #   - dist-clean - cleans release packages for all target platforms
 #   - clean-all - cleans the build area and release packages
 #   - docker-thirdparty - pulls thirdparty images (postgres)
+#   - gotools - Installs go tools, such as: golint, goimports, gocov
 
 PROJECT_NAME = fabric-ca
 ALPINE_VER ?= 3.10
@@ -61,9 +62,7 @@ PKGNAME = github.com/hyperledger/$(PROJECT_NAME)
 
 METADATA_VAR = Version=$(PROJECT_VERSION)
 
-# hard-code GO_VER until CI supports Go 1.12.9
-#GO_VER = $(shell grep "GO_VER" ci.properties |cut -d'=' -f2-)
-GO_VER ?= 1.12.9
+GO_VER = $(shell grep "GO_VER" ci.properties | cut -d '=' -f2-)
 GO_SOURCE := $(shell find . -name '*.go')
 GO_LDFLAGS = $(patsubst %,-X $(PKGNAME)/lib/metadata.%,$(METADATA_VAR))
 export GO_LDFLAGS
@@ -74,10 +73,13 @@ FVTIMAGE = $(PROJECT_NAME)-fvt
 RELEASE_PLATFORMS = linux-amd64 darwin-amd64 linux-ppc64le linux-s390x windows-amd64
 RELEASE_PKGS = fabric-ca-client
 
+TOOLS = build/tools
+
 path-map.fabric-ca-client := cmd/fabric-ca-client
 path-map.fabric-ca-server := cmd/fabric-ca-server
 
 include docker-env.mk
+include gotools.mk
 
 all: rename docker unit-tests
 
@@ -105,10 +107,10 @@ license: .FORCE
 format: .FORCE
 	@scripts/check_format
 
-imports: .FORCE
+imports: $(TOOLS)/goimports
 	@scripts/check_imports
 
-lint: .FORCE
+lint: $(TOOLS)/golint
 	@scripts/check_lint
 
 vet: .FORCE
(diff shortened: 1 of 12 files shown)


Makefile
```diff
@@ -105,10 +107,10 @@ license: .FORCE
 format: .FORCE
```
sykesm: Just FYI, you didn’t start it but you should probably understand the difference between .FORCE/FORCE and .PHONY.

https://www.gnu.org/software/make/manual/html_node/Force-Targets.html

sykesm: Also, the dependency should really be on the binary and the binary target should have a dependency on tools.go so it gets built if the binary doesn’t exist or is older than tools.go.

scripts/install_gotool.sh
```diff
@@ -0,0 +1,20 @@
...
+# for installing a pinned dependency from the tools/tools.go file.
+# It accepts two arguments:
+#
+# Arg1: the name of the tool, i.e. `goimports`
+# Arg2: the fully qualified GitHub URL, i.e., `golang.org/x/tools/cmd/goimports`
+#
+# This script is not meant to be utilized independently, the installation
+# of Go tools should be performed via the make process
+
+location=$(command -v ${1})
```
sykesm: We should really be looking at the location where our process places the tools. In most cases that will be $GOBIN but it could just as easily be build/tools or similar.

Basically, relying on the path means we’re not using the tool we think we should be using.

scripts/install_gotool.sh
```diff
@@ -0,0 +1,20 @@
...
+if [[ ! -z "${location}" ]]; then
+    echo ${1} already installed at ${location}
+    exit 0
+fi
+
+echo Installing ${1} from ${2}
+export GO111MODULE=on && cd tools && go install ${2}
```
sykesm: We can specify where to put the binary by setting GOBIN or using the -o flag.

scripts/install_gotool.sh
mastersingh24: @sykesm  I assume you are suggesting we select the spot here (and I suppose we'll need to add that to our PATH here as well)?  Probably best to use our own folder in case people have other versions of the tool they prefer already in $GOBIN?

Makefile
```diff
@@ -105,10 +106,10 @@ license: .FORCE
...
 	@scripts/check_format
 
-imports: .FORCE
+imports: .build/tools/goimports
```
sykesm: Where did `.build` come from? The other targets (e.g. `docker`) are using a `build` directory.

may also be worth having a macro setup for it instead of duplicating it.

scripts/install_gotool.sh
```diff
@@ -0,0 +1,15 @@
...
+# for installing a pinned dependency from the tools/tools.go file.
+# It accepts two arguments:
+#
+# Arg1: the name of the tool, i.e. `goimports`
+# Arg2: the fully qualified package URL, i.e., `golang.org/x/tools/cmd/goimports`
+#
+# This script is not meant to be utilized independently, the installation
+# of Go tools should be performed via the make process
+
+INSTALL_PATH=$(git rev-parse --show-toplevel)/.build/tools
```
sykesm: Why are we using git instead of letting make set the output directory?

gotools.mk
```diff
@@ -0,0 +1,14 @@
+# Copyright the Hyperledger Fabric contributors. All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+GOTOOLS = $(shell grep '_' tools/tools.go | cut -d\" -f2 | rev | cut -d/ -f1 | rev)
```
sykesm: This is clever but pretty fragile. It's looking for any line with an underscore in it. This is assuming we're only going to get package imports but you never know ...

Why not just list the targets explicitly? The idea of tools.go is to enable go.mod to version the deps; it's not really meant to be a tools manifest.

Makefile
```diff
@@ -61,9 +62,8 @@ PKGNAME = github.com/hyperledger/$(PROJECT_NAME)
 
 METADATA_VAR = Version=$(PROJECT_VERSION)
 
-# hard-code GO_VER until CI supports Go 1.12.9
-#GO_VER = $(shell grep "GO_VER" ci.properties |cut -d'=' -f2-)
-GO_VER ?= 1.12.9
+# Unit tests fail when upgrading to Go 1.12.9
```
sykesm: There's a PR to address that so we should get that PR in and remove this.

Seems like ci.properties may also be able to go away in a different commit/PR

btl5037: Since ci.properties isn't used by AZP I've updated the value in there so I can remove the hardcoded value and this comment. I'll do another PR with the removal of ci.properties and once your PR gets in, I will update CI to 1.12.9

gotools.mk
```diff
@@ -0,0 +1,14 @@
...
+
+.PHONY: gotools
+gotools: $(patsubst %,.build/tools/%, $(GOTOOLS))
+
+.build/tools/%: tools/go.mod tools/tools.go
+	@mkdir -p $(@D)
+	@$(eval TOOL = ${subst .build/tools/,,${@}})
+	@$(eval FQP = $(shell grep "_" tools/tools.go | cut -d "_" -f2 | grep ${TOOL}\"$))
```
btl5037: Do you have an issue with grepping the URL out of the tools.go file? Otherwise I need to maintain a map like we have in the Fabric repo of the URL's, and that's really what I was trying to avoid with grepping the tools out of the tools.go file above

Makefile
btl5037: The other PR was merged I will integrate those changes

gotools.mk
```diff
@@ -0,0 +1,15 @@
...
+
+GOTOOLS = gocov gocov-xml goimports golint
+
+.PHONY: gotools
+gotools: $(patsubst %,build/tools/%, $(GOTOOLS))
+
+build/tools/%: tools/go.mod tools/tools.go
+	@mkdir -p $(@D)
+	@$(eval TOOL = ${subst build/tools/,,${@}})
+	@$(eval FQP = $(shell grep ${TOOL} tools/tools.go | cut -d " " -f2 | grep ${TOOL}\"$))
```
btl5037: @sykesm Do you have an issue with grepping the URL out of the tools.go file? Otherwise I need to maintain a map like we have in the Fabric repo of the URL's, and that's really what I was trying to avoid with grepping the tools out of the tools.go file above

ci.properties
```diff
@@ -4,8 +4,8 @@ FAB_BASE_VERSION=2.0.0
 FAB_BASEIMAGE_VERSION=0.4.15
 # Pull below list of images from Hyperledger Docker Hub
 FAB_THIRDPARTY_IMAGES_LIST=kafka zookeeper couchdb
-# Set compaitable go version
-GO_VER=1.12.5
+# Set compatible go version
+GO_VER=1.12.9
```
sykesm: This doesn't hold up the PR but, is this still needed at all?

gotools.mk
sykesm: s/issue/concern/g - yes.

ci.properties
btl5037: I'm not sure, I'm going to look at the daily and release jobs and see where we might use it there. I'm working to see where these vars might be referenced in existing CI

PR #32, opened 2019-10-02
[FABC-865] Fix setting TLS files by cert/key.file

naonishijima:
This CR allows operations.tls.{cert, key}.file key to set
operations TLS certificate and key file.

It is set operation TLS files to operations.tls.{cert, key}.file key
in config. However, operations TLS struct accepts
operations.tls.{certfile, keyfile}. As a result, it doesn't not
working even if it was set in operations.tls.{cert, key}.file.

Signed-off-by: Nao Nishijima <nao.nishijima.xt@hitachi.com>

### DIFF
--- a/cmd/fabric-ca-server/config.go
+++ b/cmd/fabric-ca-server/config.go
@@ -576,6 +576,28 @@ func (s *ServerCmd) configInit() (err error) {
 		return err
 	}
 
+	// Read operations tls files
+	if s.myViper.GetBool("operations.tls.enabled") {
+		cf := s.myViper.GetString("operations.tls.cert.file")
+		if !filepath.IsAbs(cf) {
+			cf = filepath.Join(s.homeDirectory, cf)
+		}
+		if util.FileExists(cf) {
+			s.cfg.Operations.TLS.CertFile = cf
+		} else {
+			return errors.Errorf("failed to read certificate file: %s", cf)
+		}
+		kf := s.myViper.GetString("operations.tls.key.file")
+		if !filepath.IsAbs(kf) {
+			kf = filepath.Join(s.homeDirectory, kf)
+		}
+		if util.FileExists(kf) {
+			s.cfg.Operations.TLS.KeyFile = kf
+		} else {
+			return errors.Errorf("failed to read key file: %s", kf)
+		}
+	}
+
 	// The pathlength field controls how deep the CA hierarchy when requesting
 	// certificates. If it is explicitly set to 0, set the PathLenZero field to
 	// true as CFSSL expects.
--- a/cmd/fabric-ca-server/main_test.go
+++ b/cmd/fabric-ca-server/main_test.go
@@ -22,6 +22,7 @@ import (
 	"github.com/hyperledger/fabric-ca/lib"
 	"github.com/hyperledger/fabric-ca/lib/metadata"
 	"github.com/hyperledger/fabric-ca/util"
+	"github.com/spf13/viper"
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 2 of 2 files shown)


cmd/fabric-ca-server/config.go
```diff
@@ -576,6 +576,30 @@ func (s *ServerCmd) configInit() (err error) {
 		return err
 	}
 
+	// Read operations tls files
+	if s.myViper.GetString("operations.tls.enabled") == "true" {
```
sykesm: GetBool instead of string comparison

cmd/fabric-ca-server/config.go
```diff
@@ -576,6 +576,30 @@ func (s *ServerCmd) configInit() (err error) {
...
+		cf := s.myViper.GetString("operations.tls.cert.file")
+		if util.FileExists(cf) {
+			if filepath.IsAbs(cf) {
+				s.cfg.Operations.TLS.CertFile = cf
+			} else {
+				s.cfg.Operations.TLS.CertFile = filepath.Join(filepath.Dir(s.cfgFileName), cf)
+			}
+		} else {
+			return errors.New("Failed to read certificate file")
```
sykesm: Error messages are lower case; messages should include context

https://github.com/golang/go/wiki/CodeReviewComments#error-strings

cmd/fabric-ca-server/config.go
```diff
@@ -576,6 +576,30 @@ func (s *ServerCmd) configInit() (err error) {
...
 	}
 
+	// Read operations tls files
+	if s.myViper.GetString("operations.tls.enabled") == "true" {
+		cf := s.myViper.GetString("operations.tls.cert.file")
+		if util.FileExists(cf) {
+			if filepath.IsAbs(cf) {
+				s.cfg.Operations.TLS.CertFile = cf
+			} else {
+				s.cfg.Operations.TLS.CertFile = filepath.Join(filepath.Dir(s.cfgFileName), cf)
```
sykesm: This is making an assumption about the working directory. If the cert file is relative to the config path but not the current working directory, this won't work.

Honestly, I don't know why we need to determine if the file exists before trying to use it. Checking if it's absolute or relative is important, and then, when relative, that we make it relative to the config file location.

A missing file should result in a config read error.

cmd/fabric-ca-server/config.go
```diff
@@ -576,6 +576,30 @@ func (s *ServerCmd) configInit() (err error) {
...
+		if util.FileExists(kf) {
+			if filepath.IsAbs(kf) {
+				s.cfg.Operations.TLS.KeyFile = kf
+			} else {
+				s.cfg.Operations.TLS.KeyFile = filepath.Join(filepath.Dir(s.cfgFileName), kf)
+			}
+		} else {
+			return errors.New("Failed to read key file")
+		}
+	}
```
sykesm: Same as above.

cmd/fabric-ca-server/main_test.go
```diff
@@ -471,3 +473,55 @@ func getTestClient(port int, homeDir string) *lib.Client {
...
+			key:     "../../testdata/tls_server-key.pem",
+			err:     true,
+		},
+	}
+
+	for _, tt := range cases {
+		os.Setenv("FABRIC_CA_SERVER_OPERATIONS_TLS_ENABLED", tt.enabled)
+		os.Setenv("FABRIC_CA_SERVER_OPERATIONS_TLS_CERT_FILE", tt.cert)
+		os.Setenv("FABRIC_CA_SERVER_OPERATIONS_TLS_KEY_FILE", tt.key)
+		os.Setenv("FABRIC_CA_SERVER_BOOT", "user:pass")
```
sykesm: I'd prefer we avoid the env vars and just set appropriate values on the viper instance. There's a reason why these functions accept an instance of viper...

cmd/fabric-ca-server/main_test.go
```diff
@@ -471,3 +473,55 @@ func getTestClient(port int, homeDir string) *lib.Client {
...
+		var s ServerCmd
+		s.cfg = &lib.ServerConfig{}
+		s.myViper = viper.New()
+		s.myViper.SetEnvPrefix(envVarPrefix)
+		s.myViper.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
```
sykesm: This shouldn't be needed after removing the env vars.

cmd/fabric-ca-server/main_test.go
```diff
@@ -471,3 +473,55 @@ func getTestClient(port int, homeDir string) *lib.Client {
...
+			key:     "noexit.pem",
+			err:     false,
+		},
+		{
+			enabled: "false",
+			cert:    "../../testdata/tls_server-cert.pem",
+			key:     "../../testdata/tls_server-key.pem",
+			err:     true,
+		},
+	}
```
sykesm: Where are the absolute path checks?

naonishijima:
Thank you for reviewing. I updated the patch.

sykesm:
Looks good. Can you please rebase this on the current master and it's ready to merge.

Thanks!

sykesm:
@naonishijima I requested a rebase to avoid the merge. I'll fix this up, push, and merge on your behalf.

Thanks.

PR #33, opened 2019-10-02
[FABC-875] Fix Go Version in ci.props

btl5037:
This file is still sourced by the nightly CI jobs for setting the go version for the job. Needs to be reverted to an existing version until we migrate release jobs. I mistakenly changed it in my last commit.

### DIFF
--- a/ci.properties
+++ b/ci.properties
@@ -5,7 +5,7 @@ FAB_BASEIMAGE_VERSION=0.4.15
 # Pull below list of images from Hyperledger Docker Hub
 FAB_THIRDPARTY_IMAGES_LIST=kafka zookeeper couchdb
 # Set compatible go version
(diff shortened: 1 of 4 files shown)


ci.properties
```diff
@@ -5,8 +5,8 @@ FAB_BASEIMAGE_VERSION=0.4.15
 # Pull below list of images from Hyperledger Docker Hub
 FAB_THIRDPARTY_IMAGES_LIST=kafka zookeeper couchdb
 # Set compatible go version
-GO_VER=1.12.9
+GO_VER=1.12.5
```
mastersingh24: @btl5037   Should this be 1.12.9 as well?

btl5037: No, our nightly release jobs still depend on this. This is actually what this PR is fixing

PR #40, opened 2019-10-08
[FABCI-458] Make Postgres and MySQL standalone containers

btl5037:
Pulls PostgresDB and MySQL installed in the FVT container out as a standalone service container. Some of this work is middle ground on the way to a full solution, namely the run_fvt_docker script. Once I pull LDAP out in another PR I will unify the system with two make targets, one for CI with service containers, and one where we launch the containers for users running FVT locally.


scripts/fvt/db_test.sh
```diff
@@ -261,63 +261,3 @@ killserver $pid || ErrorExit "Failed to stop CA"
...
-# Enroll again, this time the server should try to reinitialize the DB before processing enroll request and this should succeed
-enroll a b 2>&1 | grep "Stored client certificate"
-if [ $? != 0 ]; then
-    ErrorMsg "Enroll request should have passed"
-fi
-
-$SCRIPTDIR/fabric-ca_setup.sh -R -x $FABRIC_CA_SERVER_HOME
-
-CleanUp $RC
-exit $RC
```
btl5037: Not sure what to do about these two tests. They stop the database service, create a new CA, start the DB, then restart the server to verify DB initialization. Not possible for us to start and stop service containers

scripts/run_fvt_docker
```diff
@@ -0,0 +1,38 @@
...
+  rm -rf .mysql_data
+}
+
+cleanup
+docker network create fvtnet
+docker run --network fvtnet -d --name ${POSTGRES_DOCKER_HOST} -e POSTGRES_PASSWORD=postgres postgres:${PG_VER}
+docker run --network fvtnet -d --name ${MYSQL_DOCKER_HOST} -v $(pwd)/${MYSQL_DOCKER_CERTS}:/var/lib/mysql -e MYSQL_ROOT_PASSWORD=mysql mysql:${MYSQL_VER}
+docker run --network fvtnet --name ${FABRIC_FVT_HOST} -v $(pwd):/opt/gopath/src/github.com/hyperledger/fabric-ca hyperledger/fabric-ca-fvt
+
+trap cleanup EXIT
```
btl5037: This will eventually become the standard file targeted by the fvt-tests make target, for users who run locally. We will have a second target `fvt-tests-ci` that will implement service containers and not use this file at all

PR #42, opened 2019-10-09
[FABC-865] Fix setting TLS files by cert/key.file

sykesm:
This CR allows operations.tls.{cert, key}.file key to set operations TLS certificate and key file.

It is set operation TLS files to operations.tls.{cert, key}.file key in config. However, operations TLS struct accepts operations.tls.{certfile, keyfile}. As a result, it doesn't not working even if it was set in operations.tls.{cert, key}.file.

(cherry picked from commit df3294560368bf120cb8373a9e72b949593b5cc0)

### DIFF
--- a/cmd/fabric-ca-server/config.go
+++ b/cmd/fabric-ca-server/config.go
@@ -576,6 +576,28 @@ func (s *ServerCmd) configInit() (err error) {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


sykesm:
@mastersingh24 - can you please review this cherry-pick from master?

PR #45, opened 2019-10-10
[FABC-880] Add license header to gencsr_test.sh

btl5037:


btl5037:
@sykesm I need to do it in both, release-1.4 is flagging it now

PR #51, opened 2019-10-15
Implement service containers both in CI and locally

btl5037:
Today MySQL and Postgres are installed via the apt package manger directly into the FVT image. This PR pulls them out as standalone service containers, giving us the freedom to manage their versions independent of what is available in the apt repository. This PR also consumes the MySQL and Postgres containers as Service Containers in Azure Pipelines


images/fabric-ca-fvt/payload/start.sh
```diff
@@ -8,21 +8,27 @@
 POSTGRES_PORT=5432
 MYSQL_PORT=3306
 LDAP_PORT=389
-PORTS=($POSTGRES_PORT $MYSQL_PORT $LDAP_PORT)
+PORTS=($MYSQL_PORT $LDAP_PORT)
```
sykesm: I'd expect all of these to be set with defaults:

 PGDATABASE, PGHOST, PGPORT and/or PGUSER

```
PGDATABASE=${PGDATABASE:-postgres}
PGHOST=${PGHOST:-postgres}
PGPORT=${PGPORT:-5432}
PGUSER=${PGUSER:-postgres}

These are from the psql man page. https://www.postgresql.org/docs/9.3/libpq-envars.html

File: scripts/fvt/db_test.sh [hunk 2]
```diff
@@ -127,7 +127,7 @@ EOF
    cp $MYSQLSERVERCONFIG $PGSQLSERVERCONFIG
    cp $MYSQLSERVERCONFIG2 $PGSQLSERVERCONFIG2
    sed -i "s/type: mysql/type: postgres/
-          s/datasource:.*/datasource: host=localhost port=$POSTGRES_PORT user=postgres password=postgres dbname=fabric_ca/" \
+          s/datasource:.*/datasource: host=postgres port=$POSTGRES_PORT user=postgres password=postgres dbname=fabric_ca sslmode=disable/" \
    $PGSQLSERVERCONFIG $PGSQLSERVERCONFIG2
 }
```
sykesm: You're just copying this but we could just cat the source through sed to create the target file. That gets rid of the `-i` flag and it's inconsistent behavior depending on the version of `sed` that's being used.

PR #55, opened 2019-10-28
FABC-881 Update vendored dependencies

mastersingh24:
- ensure that all deps from the fabric repo are at the
same revision (latest from release-1.4 branch)

- sync deps

Signed-off-by: Gari Singh <gari.r.singh@gmail.com>


mastersingh24:
will do master as well ... but not a simple cherry pick and the main issue was with 1.4.x

PR #57, opened 2019-10-28
FABC-881 Use earlier revision of bccsp

mastersingh24:
- need to revert an earlier commit as we need to
support plugins in the 1.4.x stream

Signed-off-by: Gari Singh <gari.r.singh@gmail.com>


mastersingh24:
@sykesm  - I had to use an earlier revision of bccsp (still from master) as we still need to support plugins in 1.4.x  :-(

PR #63, opened 2019-11-11
[FABCI-461] Create release job

btl5037:
This change adds an Azure Pipelines yaml
definition for a release pipeline. It builds the
fabric-CA binaries, builds and publishes the
fabric-ca docker images to dockerhub, and
creates a draft GitHub release that includes
the existing changlelog and binaries.

### DIFF
--- a/ci/azure-pipelines-release.yml
+++ b/ci/azure-pipelines-release.yml
@@ -1,0 +1,89 @@
+# Copyright the Hyperledger Fabric cont ributors. All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+name: RELEASE-$(Date:yyyyMMdd)$(Rev:.rrr)
+trigger: none
+pr: none
+
+variables:
+  - group: credentials
+  - name: GOPATH
+    value: $(Agent.BuildDirectory)/go
+  - name: GOVER
+    value: 1.12.12
+
+stages:
+  - stage: BuildBinaries
+    dependsOn: []
+    displayName: "Build Fabric CA Binaries"
+    jobs:
+      - job: Build
+        pool:
+          vmImage: ubuntu-16.04
+        container: golang:$(GOVER)
+        strategy:
+          matrix:
+            Linux-amd64:
+              TARGET: linux-amd64
+            MacOS-amd64:
+              TARGET: darwin-amd64
+            Windows-amd64:
+              TARGET: windows-amd64
+        steps:
+          - checkout: self
+            path: 'go/src/github.com/hyperledger/fabric-ca'
+            displayName: Checkout Fabric CA Code
+          - script: ./ci/create_binary_package.sh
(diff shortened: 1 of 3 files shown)


ci/azure-pipelines-release.yml
```diff
@@ -0,0 +1,91 @@
...
+name: RELEASE-$(Date:yyyyMMdd)$(Rev:.rrr)
+trigger: none
+pr: none
+
+variables:
+  - group: credentials
+  - name: GOPATH
+    value: $(Agent.BuildDirectory)/go
+  - name: GOVER
+    value: 1.12.12
```
sykesm: This is stale now.

ci/azure-pipelines-release.yml
```diff
@@ -0,0 +1,91 @@
...
+  - name: RELEASE
+    value: 1.4.4
```
sykesm: This seems weird to me. This value should be explicitly populated by the person doing the manual execution of the build. According to the doc, having this in the pipeline definition means it can't be changed:

> If a variable appears in the variables block of a YAML file, it is fixed and cannot be overridden at queue time.

ci/publish_docker.sh
```diff
@@ -0,0 +1,16 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+set -eu -o pipefail
+
+make docker
+docker login --username "${DOCKER_USERNAME}" --password "${DOCKER_PASSWORD}"
+docker tag "hyperledger/fabric-ca" "hyperledger/fabric-ca:amd64-${RELEASE}"
+docker push "hyperledger/fabric-ca:amd64-${RELEASE}"
+
+sudo wget -qO /usr/local/bin/manifest-tool https://github.com/estesp/manifest-tool/releases/download/v1.0.0/manifest-tool-linux-amd64
```
sykesm: why sudo? Do we need to put this into /usr/local/bin?

ci/azure-pipelines-release.yml
btl5037: https://github.com/hyperledger/fabric-ca/blob/release-1.4/ci.properties#L8

btl5037: This pipeline file is going into 1.4, when I replicate it to master I'll upgrdae

ci/publish_docker.sh
btl5037: Technically no, but then I'd have to populate the path variable in the variables section, or execute it locally

ci/azure-pipelines-release.yml
btl5037: Removed, I was aware of this, and had good intentions behind it (I originally had a comment as to how this would be addressed in the future), but I guess I don't really have an issue with this,

ci/publish_docker.sh
btl5037: Changed to executing manually

PR #72, opened 2019-11-28
Update operations_guide.rst

sungjinwang:
orderer1-org0: registration name and enrollment name are not the same.
admin-org0: registration name/password and enrollment name/password are not the same.


mastersingh24:
@sungjinwang  - you need to sign-off for your commit:   https://hyperledger-fabric.readthedocs.io/en/release-1.4/CONTRIBUTING.html#legal-stuff

PR #73, opened 2019-12-03
[FABC-780] Add support for TLS 1.3

btl5037:
Adds support for TLS 1.3, enabling users to use to connect to the server endpoint using either protocol

#### Type of change

#### Description of change

Changes the TLS protocol to support only 1.3 and adds the 1.2 protocol to the rejection list in tests

### DIFF
--- a/lib/server.go
+++ b/lib/server.go
@@ -666,7 +666,7 @@ func (s *Server) listenAndServe() (err error) {
 			ClientAuth:   clientAuth,
 			ClientCAs:    certPool,
 			MinVersion:   tls.VersionTLS12,
-			MaxVersion:   tls.VersionTLS12,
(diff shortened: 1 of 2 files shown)


lib/server.go
```diff
@@ -665,8 +665,8 @@ func (s *Server) listenAndServe() (err error) {
 			Certificates: []tls.Certificate{*cer},
 			ClientAuth:   clientAuth,
 			ClientCAs:    certPool,
-			MinVersion:   tls.VersionTLS12,
-			MaxVersion:   tls.VersionTLS12,
+			MinVersion:   tls.VersionTLS13,
+			MaxVersion:   tls.VersionTLS13,
```
mastersingh24: We still need to accept TLS 1.2 ... so min should still be TLS12

PR #81, opened 2019-12-20
[FABC-891] Create missing index on postgres db

jamesHanKey:
#### Postgres db users.id does not have index causing register delays.

- Bug fix

#### Description

Our group have tested on AWS aurora r5.4xlarge and got only 100 registers per second on a c5.12xlarge ec2 instance.

However, even though the CPU utilization for both database and instance was below 30% we were getting more than 10-sec delays for register and modify for 100 requests per second.

So we thought there was a bottleneck somewhere. We found that the select * from users where id = ? was taking more than 700ms. There we asked ourselves, is there no index in users table?

There it was. In hyperledger/fabric-ca repository, lib/server/db/postgres/postgres.go, no primary key or index was found in the users table compared to the MySQL version which explicitly had a primary key in id.


btl5037:
You'll also need to regenerate the mock, change to lib/server/db/postgres and run `go generate ./...`

lib/server/db/postgres/postgres_test.go
```diff
@@ -153,17 +153,26 @@ var _ = Describe("Postgres", func() {
 			Expect(err.Error()).Should(ContainSubstring("Failed to create Postgres tables: Error creating users table: unable to create table"))
 		})
 
-		It("returns an error if unable to create affiliations table", func() {
+		It("returns an error if unable to create users index", func() {
 			mockDB.ExecReturnsOnCall(1, nil, errors.New("unable to create table"))
```
btl5037: I think you've got the wrong order here, your error, should be call #2

btl5037:
LGTM, @mastersingh24 do you see an issue with adding a primary key to the Postgres DB, or was it just an accidental omission at design time?

sykesm:
LGTM; performing a local test before merging.

sykesm:
Looks good. Only thing left is to rebase on top of current master and then I'm happy to merge.

Thanks for the contribution!

manish-sethi:
@jamesHanKey - it appears that this PR is almost ready to merge. Can you rebase on top of current master?

PR #82, opened 2019-12-26
[FABC-892] Includes pkcs11 under bccsp section of default config files

caduellery:
#### Type of change

- Bug fix

#### Description

This fixes an issue with environment variables that are ignored by the fabric-ca-server and fabric-ca-client when they don't exist in the config file. Previous behavior prevented the use of environment variables to initialize Fabric CA integrated with an HSM.

#### Related issues

[FABC-892](https://jira.hyperledger.org/browse/FABC-892)

### DIFF
--- a/cmd/fabric-ca-client/command/config.go
+++ b/cmd/fabric-ca-client/command/config.go
@@ -186,7 +186,8 @@ caname:
 
 #############################################################################
 # BCCSP (BlockChain Crypto Service Provider) section allows to select which
-# crypto implementation library to use
+# crypto implementation library to use. Change "default:" value to match the
+# provider (SW or PKCS11).
 #############################################################################
 bccsp:
     default: SW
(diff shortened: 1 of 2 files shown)


mastersingh24:
First, thanks for the contribution.  But I'm not sure we really want to do this.  I'd prefer that perhaps we document this somewhere (basically you cannot override config with environment variables in the config does not exist).  Generally speaking, we need to stop thinking that config files and environment variables are interchangeable.

caduellery:
@mastersingh24 I'm ok with this decision. But I still wonder if it would be better to be explicit in the configuration template than to omit some options ... since it is a template ...

caduellery:
One more interested: https://github.com/hyperledger/fabric-ca/pull/83#issuecomment-584677808

caod123:
@mastersingh24 what's the consensus on this? This has been open for some time now and it's likely we should reach a decision.

sykesm:
Given the age of this PR, the fact the test checks are not passing, and the comments from @mastersingh24, I'm going to close this PR.

We do appreciate your efforts to help improve the project and do hope you continue to engage. Thanks.

PR #88, opened 2020-01-20
Sync metrics library to Fabric's 1.4.x

saad-karim:
Change-Id: I6b4cc814a2f17d8d73668b88a9ac9662de58562a
Signed-off-by: Saad Karim <skarim@us.ibm.com>

#### Type of change

#### Description

The metrics library in 1.4.x CA did not match the metrics library that was in 1.4.x of Fabric. 
The PR should sync up the metrics library.

### DIFF
--- a/vendor/vendor.json
+++ b/vendor/vendor.json
@@ -692,68 +692,54 @@
 			"versionExact": "master"
 		},
 		{
-			"checksumSHA1": "FaDuiEL1h9wdNYBG5uyjvlBDPEY=",
+			"checksumSHA1": "WOLMxJCYaKYleZvmGEClhuhM3BE=",
 			"path": "github.com/hyperledger/fabric/common/metrics",
-			"revision": "7891c890ab6ca541b2184a32f52d05f2f964adcf",
-			"revisionTime": "2019-10-28T13:05:25Z",
+			"revision": "7917a40ff6f694b34c971b1f8285de8b1ac39777",
+			"revisionTime": "2019-11-13T12:45:15Z",
 			"version": "master",
 			"versionExact": "master"
 		},
 		{
 			"checksumSHA1": "XA7iTPuTIO7aWvVfCb1zdSbfYn4=",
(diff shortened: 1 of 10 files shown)


vendor/vendor.json
```diff
@@ -692,68 +692,54 @@
 			"versionExact": "master"
 		},
 		{
-			"checksumSHA1": "FaDuiEL1h9wdNYBG5uyjvlBDPEY=",
+			"checksumSHA1": "WOLMxJCYaKYleZvmGEClhuhM3BE=",
 			"path": "github.com/hyperledger/fabric/common/metrics",
-			"revision": "7891c890ab6ca541b2184a32f52d05f2f964adcf",
-			"revisionTime": "2019-10-28T13:05:25Z",
+			"revision": "7917a40ff6f694b34c971b1f8285de8b1ac39777",
```
mastersingh24: @saad-karim  - why are we doing this?   I had previously updated the bccsp dependency to use a revision from the master branch so that we could use the latest from metrics on the master branch as well ... I don't think we need/want this change

saad-karim: @mastersingh24 The reasoning behind this is that in ibp-operator when we vendor 1.4.x Fabric, we run into issues when vendoring 1.4.x CA. The issue being that in 1.4.x Fabric there is no field `LabelHelp` as part of the metric structs. We want to vendor the same version of CA and Fabric ,so that is why we are trying to sync up the versions.

mastersingh24: @saad-karim 
But fabric-ca 1.4 also vendors bccsp from fabric master as well .... we specifically made that change because we were vendoring metrics from fabric master as well for the LabelHelp fix which was not backported.   I think there is much more to consider here

saad-karim: Right, fabric-ca 1.4 does vendor bccsp from fabric master. This means that when we vendor dependencies in operator we have 1.4 fabric-ca and then master fabric. Ideally, we wanted to vendor the same version of fabric and fabric-ca.

saad-karim:
Based on discussion with Gari, closing this request

PR #96, opened 2020-02-22
Fix table in user guide (currently not rendered in html due to this)

barney2k7:
Signed-off-by: barney2k7 <barney2k7@users.noreply.github.com>

#### Type of change

- Documentation update

#### Description

The table attributes that can be registered for an identity in the user guide is currently not rendered in html due to invalid syntax.
This PR fixes the broken table


barney2k7:
Note that you Azure Pipeline DocBuild, step 'Build Documentation' even [contains a corresponding error](https://dev.azure.com/Hyperledger/1cf7e7c6-8faa-4fc6-9cd2-34004758dd38/_apis/build/builds/7772/logs/20) that unfortunately didn't let that build fail.

PR #105, opened 2020-02-25
Update release make targets

btl5037:
Moves the packaging logic to the makefile and properly formats the tarballs


Makefile
```diff
@@ -272,18 +272,18 @@ release/%/bin/fabric-ca-server: $(GO_SOURCE)
 
 .PHONY: dist
 dist: dist-clean release
-	cd release/$(MARCH) && tar -czvf hyperledger-fabric-ca-$(MARCH).$(PROJECT_VERSION).tar.gz *
+	cd release/$(MARCH) && tar -czvf hyperledger-fabric-ca-$(MARCH)-$(PROJECT_VERSION).tar.gz *
 dist-all: dist-clean release-all $(patsubst %,dist/%, $(RELEASE_PLATFORMS))
```
mastersingh24: Do we still need `release-all` here?  Not that it hurts

btl5037: I have on my list to revisit the entire process in both Fabric and CA and simplify it all. Release-all just builds the binaries, then dist-all packages all of them into a tarball, so it's still necessary, despite the fact we don't personally use it anywhere. I'll remove it in the future when I get around to cleaning the targets up in both projects

PR #117, opened 2020-03-28
Updated BCCSP Vendor to include cleanup and fixes

ahmedsajid:
#### Type of change

- Bug fix

#### Description

Updated BCCSP Vendor

Signed-off-by: Ahmed Sajid <ahmed.sajid@securekey.com>


btl5037:
@mastersingh24 We removed RSA support from BCCSP [here](https://github.com/hyperledger/fabric/commit/80a20e3aaf5fcfa70e2e3cd8a3e10eb31b1d5901#diff-06abcf42f6dca2011c579f9e9df4005a) since Fabric does not use it. Is the right things to do here to remove RSA support from [here](https://github.com/hyperledger/fabric-ca/blob/master/util/csp.go) or does this have other implications?

btl5037:
We are working through some ideas on how to address this. Additional work needs to be done to make this work

sykesm:
Thank you for the contribution but I believe this has been (or will be) superseded by one of #120, #122, #123, or #124.

PR #118, opened 2020-04-01
[FABC-902] Repair the sql error,It should have two args

bjzhang03:
Signed-off-by: bjzhang <bjzhang1991@gmail.com>

#### Type of change

- Bug fix

#### Description
When I want to get the identity,I got the error,And I found the reason is the sql need two args, but we do not supply it, SO I repair it.

### DIFF
--- a/lib/dbaccessor.go
+++ b/lib/dbaccessor.go
@@ -591,7 +591,7 @@ func (d *Accessor) GetFilteredUsers(affiliation, types string) (*sqlx.Rows, erro
 	subAffiliation := affiliation + ".%"
 	if util.ListContains(types, "*") { // If type is '*', allowed to get back of all types for requested affiliation
 		query := "SELECT * FROM users WHERE ((affiliation = ?) OR (affiliation LIKE ?))"
-		rows, err := d.db.Queryx("GetFilteredUsers", d.db.Rebind(query))
+		rows, err := d.db.Queryx("GetFilteredUsers", d.db.Rebind(query), affiliation, subAffiliation)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


btl5037:
@MHBauer is correct, there is no coverage on this query, it would be great if you could add a test to cover this scenario:

image

bjzhang03:
> In my limited understanding, the code seems fine, but is there a test? If this is an error, it looks like this should have failed 14 months ago at the last change, maybe even before that.

I have added the unit test

lib/dbaccessor.go
```diff
@@ -591,7 +591,7 @@ func (d *Accessor) GetFilteredUsers(affiliation, types string) (*sqlx.Rows, erro
 	subAffiliation := affiliation + ".%"
 	if util.ListContains(types, "*") { // If type is '*', allowed to get back of all types for requested affiliation
 		query := "SELECT * FROM users WHERE ((affiliation = ?) OR (affiliation LIKE ?))"
-		rows, err := d.db.Queryx("GetFilteredUsers", d.db.Rebind(query))
+		rows, err := d.db.Queryx("GetFilteredUsers", d.db.Rebind(query), affiliation, subAffiliation)
 		if err != nil {
 			return nil, errors.Wrapf(err, "Failed to execute query '%s' for affiliation '%s' and types '%s'", query, affiliation, types)
```
sykesm, line 596: nit: can we update the error message to reflect what's being done in the query? This query does not include types but it does include the subAffiliation.

bjzhang03: I have updated the error msg and backfill testing for the other paths in GetFilteredUsers

PR #119, opened 2020-04-08
[FABC-903] Add GetFilteredUsers Unit test

bjzhang03:
add the unit test to cover GetFilteredUsers case and update the error msg

Signed-off-by: bjzhang <bjzhang1991@gmail.com>

#### Type of change

- Test update

#### Description

The getFilteredUser does not have full test path of all cases, I add the unit to fix this problem.
And update the error msg to tell the user what the query do.

#### Related issues
[FABC-903]

### DIFF
--- a/lib/dbaccessor.go
+++ b/lib/dbaccessor.go
@@ -593,7 +593,7 @@ func (d *Accessor) GetFilteredUsers(affiliation, types string) (*sqlx.Rows, erro
 		query := "SELECT * FROM users WHERE ((affiliation = ?) OR (affiliation LIKE ?))"
 		rows, err := d.db.Queryx("GetFilteredUsers", d.db.Rebind(query), affiliation, subAffiliation)
 		if err != nil {
-			return nil, errors.Wrapf(err, "Failed to execute query '%s' for affiliation '%s' and types '%s'", query, affiliation, types)
+			return nil, errors.Wrapf(err, "Failed to execute query '%s' for affiliation '%s', subAffiliation '%s',  and types '%s'", query, affiliation, subAffiliation, types)
 		}
 		return rows, nil
 	}
--- a/lib/dbaccessor_test.go
+++ b/lib/dbaccessor_test.go
@@ -287,18 +287,11 @@ func testInsertAndGetUser(ta TestAccessor, t *testing.T) {
 func testInsertAndGetFilteredUsers(ta TestAccessor, t *testing.T) {
 	t.Log("TestInsertAndGetFilteredUsers")
 	ta.Truncate()
-	// test type=* and affiliation is not nil
-	// insert affiliation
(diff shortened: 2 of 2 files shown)


lib/dbaccessor.go
```diff
@@ -593,7 +593,7 @@ func (d *Accessor) GetFilteredUsers(affiliation, types string) (*sqlx.Rows, erro
 		query := "SELECT * FROM users WHERE ((affiliation = ?) OR (affiliation LIKE ?))"
 		rows, err := d.db.Queryx("GetFilteredUsers", d.db.Rebind(query), affiliation, subAffiliation)
 		if err != nil {
-			return nil, errors.Wrapf(err, "Failed to execute query '%s' for affiliation '%s' and types '%s'", query, affiliation, types)
+			return nil, errors.Wrapf(err, "Failed to execute query '%s' for affiliation '%s',subAffiliation '%s',  and types '%s'", query, affiliation, subAffiliation, types)
```
sykesm, line 596: nit: should be a space after the comma and before `subAffiliation`.

lib/dbaccessor_test.go
```diff
@@ -287,14 +287,63 @@ func testInsertAndGetUser(ta TestAccessor, t *testing.T) {
...
+				Value: "*",
+			},
+			api.Attribute{
+				Name:  "xyz",
+				Value: "xyz",
+			},
+		},
+	}
+	// insert user
+	err := ta.Accessor.InsertUser(&insertTypesClient)
```
sykesm, line 315: Errors can't be ignored; this needs an assertion to ensure an error did not occur.

lib/dbaccessor_test.go
```diff
@@ -287,14 +287,63 @@ func testInsertAndGetUser(ta TestAccessor, t *testing.T) {
 func testInsertAndGetFilteredUsers(ta TestAccessor, t *testing.T) {
 	t.Log("TestInsertAndGetFilteredUsers")
 	ta.Truncate()
+	// create user
```
sykesm, line 290: This test seems to have a lot of duplicate code. I think structuring this with common setup, subtests (or a table driven structure), and a simple function iterate over rows with a filter function would DRY this up and make it easier to read and manage.

Basically, I see insert users from a struct literal (where the attributes don't seem to change but type and affiliation do), insert affiliations (0 or 1 right now), call get filtered users, assert the result set contains the expected data.

bjzhang03: I have removed the duplicate code.

lib/dbaccessor_test.go
bjzhang03: I have handled the error.

lib/dbaccessor.go
bjzhang03: I have done.

bjzhang03:
I have solved the problem

PR #120, opened 2020-04-08
Copy bccsp into fabric-ca tree

mastersingh24:
#### Type of change

- Improvement (tree structure)

#### Description

bccsp is not actually designed to be a consumable package
outside of the fabric src tree.  We also really do not properly
maintain the dependency within fabric-ca and we plan to
pare it down to what is needed in fabric in any case.
Going forward in fabric-ca, we will also implement the
bare minimu of crypto needed.

Given all of this, we stop vendoring bccsp and simply copy
into the fabric-ca tree.  This is based on https://github.com/hyperledger/fabric/tree/e37aa35178fdb83228f63ef4670c1a67896f081b/bccsp.
This commit was chosen to maintain support for bccsp plugins which are support in 1.4.x releases but no longer supported in subsequent releases.


sykesm:
Another ask: Can you please clearly indicate in the commit message (or code, script, readme, something) the revision (branch and commit) of bccsp that you chose as your base so it's easier to compare in the future? Thanks.

internal/pkg/bccsp/factory/test-plugin/plugin.go
```diff
@@ -8,7 +8,7 @@ package main
 import (
 	"hash"
 
-	"github.com/hyperledger/fabric/bccsp"
+	"github.com/hyperledger/fabric-ca/internal/pkg/bccsp"
```
sykesm, line 11: This is going to be problematic for plugins.

mastersingh24: Ah ... right ... I moved it into internal in master but it needs to be external here.

mastersingh24:
> Without digging in, a couple of high level questions.
> 
> 1. Why are we doing this in `release-1.4` before master? We conventionally work in master and roll things back as necessary.
> 2. Do we need to pull in all of the bccsp/idemix packages? I do not believe that the CA uses them.
> 
> ```
> $ go list -f '{{ join .Deps  "\n" }}' ./...  | sort | uniq | grep bccsp
> github.com/hyperledger/fabric-ca/vendor/github.com/hyperledger/fabric/bccsp
> github.com/hyperledger/fabric-ca/vendor/github.com/hyperledger/fabric/bccsp/factory
> github.com/hyperledger/fabric-ca/vendor/github.com/hyperledger/fabric/bccsp/signer
> github.com/hyperledger/fabric-ca/vendor/github.com/hyperledger/fabric/bccsp/sw
> github.com/hyperledger/fabric-ca/vendor/github.com/hyperledger/fabric/bccsp/utils
> ```

I have a change ready for master, but it's different than the change for release-1.4 which does not **yet** suffer from the problem of being pinned to a specific commit of bccsp.

Will look into how much of the idemix can be removed in a future change ... fabric-ca does use idemix though.

internal/pkg/bccsp/factory/test-plugin/plugin.go
mastersingh24: will move this under pkg/bccsp for now

sykesm:
> fabric-ca does use idemix though.

Yes, but does it use bccsp/idemix?

mastersingh24:
> Another ask: Can you please clearly indicate in the commit message (or code, script, readme, something) the revision (branch and commit) of bccsp that you chose as your base so it's easier to compare in the future? Thanks.

Yeah - will add to commit as well as readme within the folder

sykesm:
If we do #124, do we still need this?

mastersingh24:
will not do this in this release stream

PR #121, opened 2020-04-11
[FABC-904] Add Version Endpoint

btl5037:
Add a version endpoint to the operations server which returns the current version of the CA

### DIFF
--- a/lib/server/operations/system.go
+++ b/lib/server/operations/system.go
@@ -9,6 +9,8 @@ package operations
 import (
 	"context"
 	"crypto/tls"
+	"fmt"
+	"io"
 	"net"
 	"net/http"
 	"strings"
@@ -17,6 +19,7 @@ import (
 	"github.com/cloudflare/cfssl/log"
 	kitstatsd "github.com/go-kit/kit/metrics/statsd"
 	"github.com/gorilla/mux"
+	"github.com/hyperledger/fabric-ca/lib/metadata"
 	"github.com/hyperledger/fabric-lib-go/healthz"
 	"github.com/hyperledger/fabric/common/metrics"
 	"github.com/hyperledger/fabric/common/metrics/disabled"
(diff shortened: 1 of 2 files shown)


lib/server/operations/version.go
```diff
@@ -0,0 +1,50 @@
...
+	if err != nil {
+		logger := flogging.MustGetLogger("operations.runner")
+		logger.Errorw("failed to encode payload", "error", err)
+		resp.WriteHeader(http.StatusInternalServerError)
+	} else {
+		resp.WriteHeader(code)
+		resp.Header().Set("Content-Type", "application/json")
+		resp.Write(js)
+	}
+}
```
mastersingh24, line 50: Given you use 
```
s.mux.HandleFunc("/version", func(w http.ResponseWriter, _ *http.Request) 
```
in initializeVersionInfoHandler(), seems like you actually only need the type and no member functions?

btl5037: Ah, thanks, bad copy from Fabric where there exists a better handler mechanism

btl5037: Fixed to use the actual Handler I created.

mastersingh24:
@btl5037  have a look at my comment

btl5037:
> @btl5037 have a look at my comment

I did, `HandleFunc` takes an anonymous function, which is what I was using because I failed to realize my VersionInfoHandler implemented the `http.Handle` interface. In Fabric we don't use the `HandleFunc` handler, we use the `Handle` handler, which is what I switched to so Fabric and CA use the same methodology. `Handle` takes an object that implemets the `ServeHTTP` function on the `Handler` interface.

In my opinion, it's much cleaner to just throw out the entire `VersionInfoHandler` and just call `HandleFunc` and `io.Write` directly, but I didn't want to break from the pattern we use in Fabric. If you are OK with using `HandleFunc` I'll go back and remove the `VersionInfoHandler`'s implementation of the `Handle` interface and just use the much simpler `HandleFunc`

btl5037:
I decided to go back and use `HandleFunc` as its a much simpler implementation and has no error path

mastersingh24:
> I decided to go back and use `HandleFunc` as its a much simpler implementation and has no error path

Good thinking.  Cleaner and simpler given the task at hand

PR #124, opened 2020-04-14
Cleanup vendor.json and bring in bccsp from 1.4

sykesm:
Simple cleanup related to vendoring:
- The vendor/vendor.json file has a bunch of nonsense in it so remove those.
- Vendor the version of bccsp from release-1.4.
- Add test generated test asset to .gitignore
- Sort the items in .gitignore
- Ensure `all` is the default make target

Since bccsp@master includes breaking changes, pin to the v1 stream.


btl5037:
Can you force push to this to do a hard reset of CI, I can't retrigger it all, not even from Azure where I can usually force it by triggering on the refspec

PR #128, opened 2020-04-14
Add metrics doc generation to docs make target

sykesm:
The metrics docs weren't integrated into the doc generation targets...


sykesm:
> @sykesm Do you need to remove https://github.com/hyperledger/fabric-ca/blob/master/Makefile#L304 ?

It's still a valid target but I can remove it.

mastersingh24:
> > @sykesm Do you need to remove https://github.com/hyperledger/fabric-ca/blob/master/Makefile#L304 ?
> 
> It's still a valid target but I can remove it.

I think you removed in the PR for release-1.4 ... that was the reason I asked.

PR #134, opened 2020-04-20
Clarification about TLS CA topology

pamandrejko:
Signed-off-by: pama-ibm <pama@ibm.com>

Feedback from readers is that it was unclear that in a production env each org would have its own TLS CA. I updated the note in the TLS CA topology to  make that explicit.

#### Type of change

- Documentation update


btl5037:
@Mergifyio refresh

pamandrejko:

pamandrejko:
The merifyio request failed because this topic does not exist in the release-1.4 branch.

PR #142, opened 2020-04-22
Move AZP file to correct place

btl5037:
Signed-off-by: Ry Jones <ry@linux.com>


btl5037:
@ryjones Once this one gets in, I'll change the branch settings back to what you had, and reupdate the AZP settings to what you had

PR #147, opened 2020-05-07
Add support for .md files and variable replacement in /docs

btl5037:
This change adds support for building .md files with Sphinx.
It also adds the functionality to use placeholders so they
can be used instead of having to hardcode things like branch
names in the doc.


btl5037:

PR #152, opened 2020-05-14
[FABC-800] Enlarge PEM column for mysql db from 4096 to 8192

caod123:
#### Type of change

- Bug fix

#### Description

Also update migrator

#### Related issues

[FABC-800](https://jira.hyperledger.org/browse/FABC-800)

/cc @adecaro

### DIFF
--- a/lib/server/db/mysql/migrator.go
+++ b/lib/server/db/mysql/migrator.go
@@ -91,7 +91,8 @@ func (m *Migrator) MigrateCertificatesTable() error {
 	tx := m.Tx
 	const funcName = "MigrateCertificatesTable"
 	// Future schema updates should add to the logic below to handle other levels
-	if m.CurLevels.Certificate < 1 {
+	curLevel := m.CurLevels.Certificate
+	if curLevel < 1 {
 		log.Debug("Upgrade certificates table to level 1")
 		_, err := tx.Exec(funcName, "ALTER TABLE certificates ADD COLUMN level INTEGER DEFAULT 0 AFTER pem")
 		if err != nil {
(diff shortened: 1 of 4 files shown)


caod123:
Not sure if this also needs to be updated in the migrator

scripts/fvt/dbmigration_test.sh
```diff
@@ -119,7 +119,7 @@ mysql --host=localhost --user=root --password=mysql -e "drop database $DBNAME;"
...
 mysql --host=localhost --user=root --password=mysql --database=$DBNAME -e "CREATE TABLE affiliations (name VARCHAR(64) NOT NULL UNIQUE, prekey VARCHAR(64));"
-mysql --host=localhost --user=root --password=mysql --database=$DBNAME -e "CREATE TABLE certificates (id VARCHAR(64), serial_number varbinary(128) NOT NULL, authority_key_identifier varbinary(128) NOT NULL, ca_label varbinary(128), status varbinary(128) NOT NULL, reason int, expiry timestamp DEFAULT 0, revoked_at timestamp DEFAULT 0, pem varbinary(4096) NOT NULL, PRIMARY KEY(serial_number, authority_key_identifier)) DEFAULT CHARSET=utf8 COLLATE utf8_bin;"
+mysql --host=localhost --user=root --password=mysql --database=$DBNAME -e "CREATE TABLE certificates (id VARCHAR(64), serial_number varbinary(128) NOT NULL, authority_key_identifier varbinary(128) NOT NULL, ca_label varbinary(128), status varbinary(128) NOT NULL, reason int, expiry timestamp DEFAULT 0, revoked_at timestamp DEFAULT 0, pem varbinary(8192) NOT NULL, PRIMARY KEY(serial_number, authority_key_identifier)) DEFAULT CHARSET=utf8 COLLATE utf8_bin;"
```
sykesm, line 122: IIUC, This is supposed to be the pre-migration table. The code in `lib/server/db/mysql/migrator.go` needs to changed to expand the column.

Validation that the column was expanded is lower down.

caod123: Yep just realized that as I was updating the migrator

lib/server/db/mysql/migrator.go
```diff
@@ -103,6 +103,10 @@ func (m *Migrator) MigrateCertificatesTable() error {
 		if err != nil {
 			return err
 		}
+		_, err = tx.Exec(funcName, "ALTER TABLE certificates MODIFY pem varbinary(8192)")
+		if err != nil {
+			return err
+		}
```
sykesm, line 109: And this does not look like it would apply if the first migration took place, right?

caod123: good catch

PR #156, opened 2020-05-15
[FABC-909] Check if database already exists for postgres, mysql and sqlite

mirioeggmann:
#### Type of change

- Bug fix

#### Description

Fix to be able to run Fabric CA using Postgres (/MySQL) with a database user that doesn’t have the permission to create databases on the DB Server

#### Related issues

[FABC-909](https://jira.hyperledger.org/browse/FABC-909)

### DIFF
--- a/lib/server/db/factory/factory.go
+++ b/lib/server/db/factory/factory.go
@@ -22,6 +22,7 @@ import (
 type DB interface {
 	Connect() error
 	PingContext(ctx context.Context) error
+	Exists() (bool, error)
 	Create() (*db.DB, error)
 }
 
--- a/lib/server/db/mysql/mysql.go
+++ b/lib/server/db/mysql/mysql.go
@@ -8,6 +8,7 @@ package mysql
 
 import (
 	"context"
+	"fmt"
 	"regexp"
 	"strings"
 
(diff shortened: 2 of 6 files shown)


btl5037:
Thank you for the contribution!

The implementation you've provided here, unnecessarily adds the `Exists` function onto the DB interface. Since you implement all of the logic locally to each DB implementation, it simply could have been a local private function.

Instead, push the call to `Exists` up in the CA itself, and most of the logic you've implemented goes away. Simply call `Exists` prior to the CA calling `createDatabase` and check if you should create the database: https://github.com/hyperledger/fabric-ca/blob/master/lib/ca.go#L596

This will also allow you to better separate the error logic handling.

mirioeggmann:
Alright great, thank you for the advise! However I’m not entirely sure how I should implement it. Because sqlite has no ```CreateDatabase()``` function and the interface function ```Create()``` has to be called wether or not the database already exists (because of the tables).

Can you maybe elaborate on your idea?

sykesm:
Hello, @btl5037 & @mirioeggmann. It looks like this PR has been hanging out for quite a while and I can't really tell what its status is. It looks like there were discussions on the implementation but then things stalled.

Is this still something that folks are interested in getting merged?

Thanks.

sykesm:
The problem that this PR is fixing is real so it would be a valuable contribution but it's been several months since the last interaction. Given the inactivity, I'm going to close this PR. If anyone wants to pick it back up and bring it across the finish line, we'd welcome the help.

PR #157, opened 2020-05-18
Move to Go modules

mastersingh24:
#### Type of change

- Improvement (dependency management)

#### Description

Vendoring was previously accomplished using govendor
which has been sunset in favor of Go modules.

Also updated some versions of some dependencies as
well as corresponding code which exploited them.


lib/server_test.go
```diff
@@ -2881,8 +2881,8 @@ func TestPrometheusMetricsE2E(t *testing.T) {
 
 	gt.Expect(body).To(ContainSubstring(`# HELP api_request_count Number of requests made to an API`))
 	gt.Expect(body).To(ContainSubstring(`# TYPE api_request_count counter`))
-	gt.Expect(body).To(ContainSubstring(`api_request_count{api_name="enroll",ca_name="ca",status_code="201"} 1.0`))
-	gt.Expect(body).To(ContainSubstring(`api_request_count{api_name="enroll",ca_name="ca",status_code="401"} 1.0`))
+	gt.Expect(body).To(ContainSubstring(`api_request_count{api_name="enroll",ca_name="ca",status_code="201"} 1`))
```
mastersingh24, line 2884: So I'm not sure if the change to using the promhttp handler ( https://github.com/hyperledger/fabric-ca/pull/157/files#diff-183232ffcd508f973f5c4867933ffabeR136 ) caused this or if I missed something elsewhere

mastersingh24: @sykesm  - does this make sense to you?   I'll dig a bit deeper, but just wondering if I broke something or if something in the underlying prom code changed when I updated the underlying package

sykesm: It makes sense: https://github.com/prometheus/common/commit/fdd3d500fc554db8fd97bbf83be49bbd8ad88304

sykesm: Also, we should probably try to improve the tests to make them less fragile. We have this issue with the metrics in fabric as well. These assertions were expedient but not necessarily "good".

mastersingh24: agreed on improving them ... just wanted to make sure I did not break anything this time

mastersingh24:
@sykesm @btl5037   Any changes you want to see in this PR?  Else can we get this one merged so I can continue cleaning up Fabric CA?
More work to do removing gopath from current fvt as well as moving fvt away from bash scripts

scripts/run_unit_tests
```diff
@@ -22,13 +23,13 @@ EXCLUDED_PKGS=(
     "/metrics"
 )
 
-PKGS=`go list github.com/hyperledger/fabric-ca/... | grep -v -f <(printf '%s\n' "${EXCLUDED_PKGS[@]}")`
+PKGS=`go list ./... | grep -v -f <(printf '%s\n' "${EXCLUDED_PKGS[@]}")`
```
sykesm, line 26: This is assuming the current working directory is the fabric repo.

sykesm:
Expanding on my previous comment, since these changes are more related to how we vendor dependencies than they are about turning the repo into a module, I'd like to consider renaming the `go.mod` and `go.sum` to something like `vendor.mod` and `vendor.sum`. This would prevent the repo from being treated as a module while allowing us to track the dependencies. It's hacky but prevents some misunderstandings and odd behaviors.

It would be nice if this could be handled by the `-modfile` flag but, it can't. That flag requires a `go.mod` to be present in the module root. Now that go.mod can be empty, but it has to exist... Easy enough to touch the file before managing deps if we need to.

The other changes that I think we should undo are the ones where the import path was replaced with relative `./...` on commands. Regardless of which mode we're in and which directory we're in, the import path can be used on go commands. When the relative path is used (especially from scripts) we it adds implicit assumptions about where the scripts are being called from.

With that, and the assumption that there will be follow on work to actually convert this to a module, I'd be. good with merging.

Makefile
```diff
@@ -167,6 +168,15 @@ unit-test: unit-tests
 int-tests: gotools fabric-ca-server fabric-ca-client
 	@scripts/run_integration_tests
 
+vendor: .FORCE
```
mastersingh24, line 171: @sykesm  - I'm sure there's a cleaner way to do this, but this seems to work and leaves us without a `go.mod` and `go.sum`  in the root of the repo

mastersingh24:
> It would be nice if this could be handled by the `-modfile` flag but, it can't. That flag requires a `go.mod` to be present in the module root. Now that go.mod can be empty, but it has to exist... Easy enough to touch the file before managing deps if we need to.

I think I accomplished this

Makefile
```diff
@@ -167,6 +168,15 @@ unit-test: unit-tests
...
+	@rm go.mod
+	@touch go.mod
+	@mv vendor.sum go.sum
+	@go mod tidy -modfile vendor.mod
+	@go mod vendor  -modfile vendor.mod
+	@mv go.sum vendor.sum
+	@rm go.mod
+
```
sykesm, line 179: I think we're getting to the point of diminishing returns but I don't think this is quite right.

The target is doing this:
```sh
$ make -n vendor
rm go.mod
touch go.mod
mv vendor.sum go.sum
go mod tidy -modfile vendor.mod
go mod vendor  -modfile vendor.mod
mv go.sum vendor.sum
rm go.mod
```

```suggestion
vendor: .FORCE
	@echo > go.mod
	@go mod tidy -modfile vendor.mod
	@go mod vendor  -modfile vendor.mod
	@rm go.mod

```

Makefile
```diff
@@ -29,6 +29,7 @@
 #   - clean-all - cleans the build area and release packages
 #   - docker-thirdparty - pulls thirdparty images (postgres)
 #   - gotools - Installs go tools, such as: golint, goimports, gocov
+#	- vendor - vendors third-party packages
```
sykesm, line 32: nit: this has a tab instead of spaces; nbd but, consistency

Makefile
mastersingh24: @sykesm thanks ... went with this

Makefile
mastersingh24: done

5pecia1:
> I think my biggest concern with this is not the vendoring but the fact that isn't not actually moving to a module. I'm unable to clone and build locally unless I have a GOPATH set and then I'm in this weird place where I'm in a GOPATH, with code that relies on the GOPATH, but I'm in module mode.


> Expanding on my previous comment, since these changes are more related to how we vendor dependencies than they are about turning the repo into a module, I'd like to consider renaming the `go.mod` and `go.sum` to something like `vendor.mod` and `vendor.sum`. This would prevent the repo from being treated as a module while allowing us to track the dependencies. It's hacky but prevents some misunderstandings and odd behaviors.

Hello @sykesm .
Why do you know that Fabric CA doesn't want to be a module?
Do you have any issues or documents about it?
I couldn't find about this.

Thank you.

sykesm:
> Why do you know that Fabric CA doesn't want to be a module?

As maintainers, we would like all of the Go code related to Fabric to be proper modules. Given the widespread assumption about GOPATH, it's not always as simple as dropping in a go.mod and moving on.

We want to ensure that we can clone the repository and build and test without relying on the tree being part of a GOPATH. Many tests and most of the build infrastructure in this repository have latent dependencies on a proper GOPATH so "clone and go" doesn't really work.

We also have issues with test. Some examples of the issues are described [here](https://blog.golang.org/migrating-to-go-modules#TOC_4.1.).

The changes merged as part of this PR simply move away from the no longer maintained `govendor` tool. We're using `go.mod`/`go.sum` and `go mod` to manage these dependencies but we are going out of our way to rename the file as a signal to consumers that we have not done the work to make this a proper module.

5pecia1:
@sykesm 
Thanks to the detailed explanation I understood this Fabric CA module management.

Regards.

PR #159, opened 2020-05-20
[FABC-829] Add hf.AffiliationMgr and hf.GenCRL attributes to migrated registrar users

caod123:
#### Type of change

- New feature

#### Description

Since the initial release of Fabric CA, the list of reserved attributes (hf.*) has grown. A certain subset of users upon migration from an older version of CA to a newer one should be given any newly introduced attributes that have been added in the new release.

#### Related issues

[FABC-829](https://jira.hyperledger.org/browse/FABC-829)

### DIFF
--- a/lib/server/user/user.go
+++ b/lib/server/user/user.go
@@ -520,6 +520,14 @@ func (u *Impl) Migrate(tx userDB) error {
 		}
 		currentLevel++
 	}
+
+	if currentLevel < 2 {
+		err := u.migrateUserToLevel2(tx)
+		if err != nil {
+			return err
+		}
+		currentLevel++
+	}
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


caod123:
Questions for consideration:
1. This change only adds the the attributes to any registrar users. Is that enough or should they be added to all migrated users?
2. This change only adds the two attributes `hf.AffiliationMgr` and `hf.GenCRL` at request of the original JIRA issue. Is there a need to also add logic for the remaining attributes in https://github.com/hyperledger/fabric-ca/blob/master/lib/attr/attribute.go#L41-L50

caod123:
> 1. This change only adds the the attributes to any registrar users. Is that enough or should they be added to all migrated users?

I misunderstood how the attributes are added. I believe all users will need to have these attributes so they can be registered by the registrar.

mastersingh24:
> 1. This change only adds the the attributes to any registrar users. Is that enough or should they be added to all migrated users?

These should only be added to registrars as you can only register users if you have that role.


> 2. This change only adds the two attributes `hf.AffiliationMgr` and `hf.GenCRL` at request of the original JIRA issue. Is there a need to also add logic for the remaining attributes in https://github.com/hyperledger/fabric-ca/blob/master/lib/attr/attribute.go#L41-L50


As best as I can tell, I believe these might be the only roles added since the initial release

lib/server/user/user.go
```diff
@@ -548,6 +556,38 @@ func (u *Impl) migrateUserToLevel1(tx userDB) error {
...
+	// Update identity to level 2
+	_, err := u.GetAttribute("hf.AffiliationMgr") // Check if user already has "hf.AffiliationMgr" attribute
+	if err != nil {
+		newAttr := api.Attribute{Name: "hf.AffiliationMgr", Value: "true"}
+		err := u.ModifyAttributesTx(tx, []api.Attribute{newAttr})
+		if err != nil {
+			return errors.WithMessage(err, "Failed to set attribute")
+		}
+		u.attrs[newAttr.Name] = newAttr
+	}
```
mastersingh24, line 571: We really only want this for users with the `hf.Registrar.Roles` attribute

lib/server/user/user.go
```diff
@@ -548,6 +556,38 @@ func (u *Impl) migrateUserToLevel1(tx userDB) error {
...
+
+	_, err = u.GetAttribute("hf.GenCRL") // Check if user already has "hf.GenCRL" attribute
+	if err != nil {
+		newAttr := api.Attribute{Name: "hf.GenCRL", Value: "true"}
+		err := u.ModifyAttributesTx(tx, []api.Attribute{newAttr})
+		if err != nil {
+			return errors.WithMessage(err, "Failed to set attribute")
+		}
+		u.attrs[newAttr.Name] = newAttr
+	}
```
mastersingh24, line 581: We really only want this for users with the `hf.Registrar.Roles` attribute

caod123:
Makes sense, guess I had it right the first time and misunderstood.

caod123:
Reminder we'll need to backport this after it's merged.

caod123:
@mastersingh24 are you ready to merge this?

mastersingh24:
@caod123  - are you planning to backport this one?

caod123:

btl5037:

PR #163, opened 2020-05-25
Move api and util to internal/pkg

mastersingh24:
#### Type of change

- Improvement (structure, hygiene)

#### Description

fabric-ca is not currently structured or fit to be
consumed by external clients.

This change moves the api and util folders under
internal/pkg.

#### Additional details

Plan is to restructure the code in a manner similar to https://github.com/golang-standards/project-layout

We should be able to install the server and client and perhaps have a client library as well.


sykesm:
I merged this but the builds do not work locally or in CI. Gives me all sorts of warm fuzzy feelings. I'll either fix this (in the next 30) or revert.

PR #169, opened 2020-05-26
CI should be exercising integration tests

sykesm:
- Add int-tests to the all-tests target
- Cleanup (or at least improve) the README in scripts/fvt
- Remove dead make targets from the Jenkins environment


Makefile
```diff
@@ -148,18 +144,16 @@ build/image/fabric-ca-fvt/$(DUMMY):
 	@touch $@
 
 
-all-tests: gotools fabric-ca-server fabric-ca-client
-	@scripts/run_unit_tests
+all-tests: unit-tests int-tests
```
mastersingh24, line 147: cool ... this makes more sense now

PR #172, opened 2020-06-02
[FABC-913] Set a primary key to users table for SQLite

satota2:
This patch sets a primary key with `id` to `users` table for SQLite like the implementations of the other RDBs (MySQL, PostgreSQL).

The main reason is to prevent duplicate registration of the same ID depending on the timing.

#### Type of change

- Bug fix

#### Description

When using SQLite, duplicate registration of the same ID would occur depending on the timing. After that, the ID would not work. 

This is because the users table does not have the primary key `id`.
The implementations for other RDBs (MySQL, PostgreSQL) set `id` as the primary key in `users` table.

So, this patch sets a primary key with `id` to `users` table for SQLite.

#### Related issues

- https://jira.hyperledger.org/browse/FABC-913

### DIFF
--- a/lib/dbaccessor_test.go
+++ b/lib/dbaccessor_test.go
@@ -330,7 +330,7 @@ func testInsertAndGetFilteredUsers(ta TestAccessor, t *testing.T) {
 			t.Errorf("Failed to get read row! error: %s", err)
 		}
 		// get the target user
-		if id.Name == insert.Name && id.Type == "client" {
+		if id.Name == "testTypesClient" && id.Type == "client" {
 			typesClientSuccessFlag = true
 		}
 	}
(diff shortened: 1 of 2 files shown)


sykesm:
@saad-karim can you please take a look at this? The conditional table creation adds a primary key to make it consistent with MySQL and PostgreSQL. No concerns there but I'd like your opinion on migration.

1. Based on my (quick) look through the code, it looks easy enough to multiple records for a single user name without the primary key constraint.
2. Adding the constraint means that if there are multiple user names, it's a problem that requires manual intervention to address.
3. To be complete, I believe there should be a migration for this to add PK constraint. SQLite, however, does not support altering a table to add a primary key so the table would need to be renamed, a new table created, and the records populated into the new table. Also an automated migration would fail if there were duplicate IDs in the users table.

Thanks.

saad-karim:
@sykesm Right, if multiple entries exist with the repeated username than those entries would have to be cleaned up manually before migrating to the new table schema. 

Also, yes there is no easy way to update the schema for sqlite, the process you describe to update is what we did for our earlier schema updates. 

That being said, I don't think multiple entries existing on primary key would be too much of an issue. There is a check done at https://github.com/hyperledger/fabric-ca/blob/master/lib/serverregister.go#L166 that makes sure the user does not exist before registering. Two entries could still make it through but I think the chances would be low.

PR #174, opened 2020-06-18
[FABC-912] Remove label and pin from logs

stephyee:
#### Type of change

- Bug fix

#### Description

Redacts label and pin from bccsp initialization logs

#### Related issues

[FABC-912](https://jira.hyperledger.org/browse/FABC-912)

### DIFF
--- a/internal/pkg/util/configurebccsp.go
+++ b/internal/pkg/util/configurebccsp.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/cloudflare/cfssl/log"
 	"github.com/hyperledger/fabric/bccsp/factory"
+	"github.com/hyperledger/fabric/bccsp/pkcs11"
 	"github.com/pkg/errors"
 )
 
@@ -59,8 +60,16 @@ func ConfigureBCCSP(optsPtr **factory.FactoryOpts, mspDir, homeDir string) error
 		log.Debugf("Initializing BCCSP with software options %+v", opts.SwOpts)
 	}
 	if opts.Pkcs11Opts != nil {
-		log.Debugf("Initializing BCCSP with PKCS11 options %+v", opts.Pkcs11Opts)
+		log.Debugf("Initializing BCCSP with PKCS11 options %+v", sanitizePKCS11Opts(*opts.Pkcs11Opts))
 	}
 	*optsPtr = opts
 	return nil
+}
+
+// redacts label and pin from PKCS11 opts
+func sanitizePKCS11Opts(opts pkcs11.PKCS11Opts) pkcs11.PKCS11Opts {
+	mask := strings.Repeat("*", 6)
+	opts.Pin = mask
+	opts.Label = mask
+	return opts
 }
--- a/internal/pkg/util/configurebccsp_test.go
+++ b/internal/pkg/util/configurebccsp_test.go
@@ -1,0 +1,51 @@
+// +build pkcs11
+
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package util
+
+import (
+	"testing"
+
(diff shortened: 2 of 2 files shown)


internal/pkg/util/configurebccsp.go
```diff
@@ -59,8 +60,15 @@ func ConfigureBCCSP(optsPtr **factory.FactoryOpts, mspDir, homeDir string) error
...
 	if opts.Pkcs11Opts != nil {
-		log.Debugf("Initializing BCCSP with PKCS11 options %+v", opts.Pkcs11Opts)
+		log.Debugf("Initializing BCCSP with PKCS11 options %+v", sanitizePkcs11Opts(*opts.Pkcs11Opts))
 	}
 	*optsPtr = opts
 	return nil
 }
+
+// redacts label and pin from PKCS11 opts
+func sanitizePkcs11Opts(opts pkcs11.PKCS11Opts) pkcs11.PKCS11Opts {
```
caod123, line 70: can you add a quick test to make sure this works?

stephyee: It's an unexported function, only used when logging at debug level.

caod123: You could still technically test the logger output or just have the test file in the same package as util. I guess I'm indifferent either way, will leave up to the ca maintainers to approve and merge if they feel a test is unneeded.

internal/pkg/util/configurebccsp.go
```diff
@@ -59,8 +60,15 @@ func ConfigureBCCSP(optsPtr **factory.FactoryOpts, mspDir, homeDir string) error
...
+	opts.Pin = strings.Repeat("*", len(opts.Pin))
+	opts.Label = strings.Repeat("*", len(opts.Label))
```
sykesm, line 72: It's probably not a big deal but if we're trying to redact the sensitive information, let's not use the length of the fields to determine how many asterisks are used. Just put down 6 or 8 of them and move on.

internal/pkg/util/configurebccsp.go
stephyee: Sure, I added a new test file specifically for pkcs11 configuration.

sykesm: If we're doing this to fix a bug, there should be a test. If there's no test, there's no reason to keep the behavior.

On a related note, I'm pretty sure the pkcs11 tag isn't even set when running unit tests so that should make this even more fun.

sykesm:
I threw a panic into `ConfigureBCCSP` and ran the tests; nothing failed. Further evidence that tests don't run with the pkcs11 tag.

internal/pkg/util/configurebccsp_test.go
```diff
@@ -0,0 +1,61 @@
+// +build pkcs11
+
+/*
+Copyright IBM Corp. 2016 All Rights Reserved.
```
sykesm, line 4: 2016?

internal/pkg/util/configurebccsp_test.go
```diff
@@ -0,0 +1,61 @@
+// +build pkcs11
```
sykesm, line 1: Still not sure any tests are run with this tag.

stephyee: Agreed, I ran the test manually.

internal/pkg/util/configurebccsp_test.go
stephyee: oops, I created this based off an existing file.

sykesm:
Assuming #176 gets merged, I'd like to have this rebased on top of that so we can see the tests run in CI.

That work?

btl5037:
@MErgifyio backport release-1.4

btl5037:

PR #176, opened 2020-06-24
Run unit tests with and without pkcs11 tags

sykesm:
So, yeah. Let's have tests that exercise things that only get built when the `pkcs11` tag is set...

Also, removed the post-processing of test results for test coverage and forced tests serially run one package at a time because of test pollution. (Looks like port conflicts and multiple instances of SQLite running concurrently against the same directory.)


btl5037:
LGTM, though I can't approve it

PR #181, opened 2020-07-22
Move StartNonceSweeper out of NonceManager constructor

wlahti:
#### Type of change

#### Description

First commit: 
- No need to log the same thing twice, especially when the second log has less context about the error. 

Second commit:
- Move StartNonceSweeper out of NonceManager constructor, since not all uses of the NonceManager require/expect a separate go thread to be spawned to sweep for expired nonces. Have the server call it instead.

#### Related issues

[FABC-911](https://jira.hyperledger.org/browse/FABC-911)

### DIFF
--- a/lib/server.go
+++ b/lib/server.go
@@ -36,6 +36,7 @@ import (
 	"github.com/hyperledger/fabric-ca/lib/metadata"
 	"github.com/hyperledger/fabric-ca/lib/server/db"
 	dbutil "github.com/hyperledger/fabric-ca/lib/server/db/util"
+	idemix "github.com/hyperledger/fabric-ca/lib/server/idemix"
 	servermetrics "github.com/hyperledger/fabric-ca/lib/server/metrics"
 	"github.com/hyperledger/fabric-ca/lib/server/operations"
 	stls "github.com/hyperledger/fabric-ca/lib/tls"
@@ -195,6 +196,10 @@ func (s *Server) Start() (err error) {
 		return nil
 	}
 
+	for _, ca := range s.caMap {
+		startNonceSweeper(ca)
+	}
+
 	// Start listening and serving
 	err = s.listenAndServe()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


wlahti:
Just a start on addressing this JIRA. Potentially more to come.

lib/server/idemix/nonce.go
```diff
@@ -145,7 +145,6 @@ func (nm *nonceManager) startNonceSweeper() {
 func (nm *nonceManager) sweep(curTime time.Time) error {
```
btl5037, line 145: This entire function `sweep` seems useless. The two calls to it should just be replaced with a call to `removeExpiredNoncesFromDB` directly

btl5037:
There are quite a few issues here. In particular the error returned by the DB call is explicitly ignored in the production code. The interface defines a `SweepExpiredNonces` function, but that literally only used in tests. So this could probably be cleaned up even further by removing the function from the interface. Removing the implementation of `SweepExpiredNonces` and then we need to figure out if explicitly ignoring the error is actually OK, and if it is, we should log it in `removeExpiredNoncesFromDB` but never return it, since `startNonceSweeper` explicitly ignores it

lib/server.go
```diff
@@ -373,6 +373,10 @@ func (s *Server) initDefaultCA(renew bool) error {
 	if err != nil {
 		return err
 	}
+	if ca.issuer.NonceManager() != nil {
```
saad-karim, line 376: We don't want to start the nonce sweeper when we are just initializing the CA. This should be called only when we start the server, maybe the `Start` method might be the appropriate place for this. 

Also, this applies to more than just the default CA. Take a look at the method `initMultiCAConfig`, a single server can contain multiple CA so this needs to apply to all of them.

lib/server/idemix/nonce.go
```diff
@@ -49,14 +49,19 @@ type NonceManager interface {
 	CheckNonce(nonce *fp256bn.BIG) error
 	// SweepExpiredNonces removes expired nonces from the database
 	SweepExpiredNonces() error
+	// StartNonceSweeper starts a separate thread that will remove expired
+	// nonces at the interval speciifed by the idemix.noncesweepinterval. This
+	// function should be called while initializing the server.
+	StartNonceSweeper()
```
sykesm, line 55: Since this is optional (since we conditionally check for nil in the main) it's not a required part of the contract and should not be here.

wlahti:

wlahti:
> @wlahti is not allowed to run commands

Well that isn't very nice... ;)

btl5037:
> > @wlahti is not allowed to run commands
> 
> Well that isn't very nice... ;)

OWNED!

btl5037:

PR #185, opened 2020-08-12
Edits to use a CA

joealewine:
Doc previously, and in error, listed users folders
in the org MSP section where none is needed thanks
to Node OUs

Signed-off-by: joe-alewine <Joe.Alewine@ibm.com>

#### Type of change

- Documentation update

#### Description

Fixing small errors in this doc


docs/source/deployguide/use_CA.md
```diff
@@ -68,11 +68,13 @@ Even if you don't plan to create an orderer organization, this kind of structure
...
+Now it is possible, and recommended, to encode not just `peer` or `client`, but also `orderer` or `admin` roles into the certificates generated by a CA using NodeOUs. This embeds the role an identity has within the certificate, eliminating the need for a certificate to be given a role by its placement in a particular folder, we we'll see later.
+
+Note that an identity can only have one of these roles, and that to enable these roles you must copy the relevant stanzas into a file called `config.yaml`. This `config.yaml` file is used by Fabric in different ways. In a channel MSP, it is used to verify that the `admin` of an organization has a role of `admin` (this replaces the use of an `admincerts` folder which was used in older versions of Fabric). In the local MSP of a node, it is used not just to verify the `admin` role of the node admin but that the identity of the node itself has a `peer` or `orderer` role.
```
denyeart, line 77: " In the local MSP of a node, it is used not just to verify the `admin` role of the node admin but that the identity of the node itself has a `peer` or `orderer` role."

Awkward sentence... what are you really trying to convey here?

docs/source/deployguide/use_CA.md
```diff
@@ -68,11 +68,13 @@ Even if you don't plan to create an orderer organization, this kind of structure
...
 
-Now it is possible, and recommended, to encode not just `peer` or `client`, but also `orderer` or `admin` roles into the certificates generated by a CA using NodeOUs. Note that an identity can only have one of these roles, and that to enable these roles you must copy the relevant stanzas into a file called `config.yaml`. This `config.yaml` file is used by Fabric in different ways. In a channel MSP, it is used to verify that the `admin` of an organization has a role of `admin` (this replaces the use of an `admincerts` folder which was used in older versions of Fabric). In the local MSP of a node, it is used not just to verify the `admin` role of the node admin but that the identity of the node itself has a `peer` or `orderer` role.
+Now it is possible, and recommended, to encode not just `peer` or `client`, but also `orderer` or `admin` roles into the certificates generated by a CA using NodeOUs. This embeds the role an identity has within the certificate, eliminating the need for a certificate to be given a role by its placement in a particular folder, we we'll see later.
```
denyeart, line 75: "eliminating the need for a certificate to be given a role by its placement in a particular folder, we we'll see later."

Besides the obvious typo, the user won't understand this at all...maybe better not to mention it at all here and just expand on it later.

joealewine:
I want @saad-karim to review this before we merge it, as I'm still not 100% sure I've got this right.

docs/source/deployguide/use_CA.md
```diff
@@ -283,5 +284,7 @@ Where the folders and certificates are:
 
 Note that while the certificates themselves can be named anything you want, you should not change the name of the folders themselves, as Fabric expects to consume folders with certain names.
 
+Just as Node OUs make it no longer necessary to include a certificate of an admin in the organization MSP, it is not necessary to include the public certificate of a node admin to administer the node. **Any identity given a Node OU of `admin` by the CA listed in `config.yaml` can administer any of the nodes owned by that organization without needing to place the public certificate of that admin in the organization MSP or the local MSP**.
```
pamandrejko, line 287: Should we be more specific here:
Just as Node OUs make it no longer necessary to include a certificate of an admin in the organization MSP `admincerts` folder, it is not necessary to include the public certificate of a node admin to administer the node.

Also
**Any identity given a Node OU of `admin` by the CA listed in`config.yaml` `certificate` field

sykesm:
@saad-karim PTAL

docs/source/deployguide/use_CA.md
```diff
@@ -68,11 +68,13 @@ Even if you don't plan to create an orderer organization, this kind of structure
 
 It is the best practice to create these folders before enrolling identities and then referencing them when issuing the enroll command through the `--mspdir` flag. Note that while the --mspdir flag is used to specify where the MSP of the CA admin is during **registration**, it is used instead during **enrollment** to specify the location on the filesystem where the folders and certificates returned by the CA will be stored.
 
-### Specifying NodeOUs
```
mastersingh24, line 71: I'm actually confused as to why we actually have this in the CA docs to start with.
**NodeOUs** are a Fabric construct not a Fabric CA construct.

It is true that Fabric CA will include the *role* as an *OU* in the certificate.  Do you really need to say more than that?

joealewine: The goal of the doc is to be a one stop shop for users who want to register and enroll. This is the doc we're going to point people to and tell them that they have to go through before they can create a peer or an ordering service. Once people are done here hopefully all they have to do is point to the correct folders/certs.

So I get your point, but I do think this belongs here completeness' sake.

pamandrejko:

PR #189, opened 2020-09-10
Allow reenroll to reuse existing private key

wlahti:
#### Type of change

- New feature

#### Description

Allow reenroll to reuse existing private key instead of always generating a new one.

#### Related issues

[FABC-914](https://jira.hyperledger.org/browse/FABC-914)

### DIFF
--- a/internal/pkg/api/client.go
+++ b/internal/pkg/api/client.go
@@ -306,10 +306,13 @@ type TimeRange struct {
 	EndTime   string
 }
 
-// KeyRequest encapsulates size and algorithm for the key to be generated
+// KeyRequest encapsulates size and algorithm for the key to be generated.
+// If ReuseKey is set, reenrollment requests will reuse the existing private
+// key.
 type KeyRequest struct {
-	Algo string `json:"algo" yaml:"algo" help:"Specify key algorithm"`
-	Size int    `json:"size" yaml:"size" help:"Specify key size"`
+	Algo     string `json:"algo" yaml:"algo" help:"Specify key algorithm"`
+	Size     int    `json:"size" yaml:"size" help:"Specify key size"`
+	ReuseKey bool   `json:"reusekey" yaml:"reusekey" help:"Reuse existing key during reenrollment"`
 }
 
 // Attribute is a name and value pair
--- a/lib/client.go
+++ b/lib/client.go
@@ -8,6 +8,8 @@ package lib
 
 import (
 	"bytes"
+	"crypto"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
@@ -33,6 +35,7 @@ import (
 	"github.com/hyperledger/fabric-ca/lib/streamer"
 	"github.com/hyperledger/fabric-ca/lib/tls"
 	"github.com/hyperledger/fabric/bccsp"
+	cspsigner "github.com/hyperledger/fabric/bccsp/signer"
 	"github.com/hyperledger/fabric/idemix"
 	"github.com/mitchellh/mapstructure"
 	"github.com/pkg/errors"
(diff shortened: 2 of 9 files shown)


yeasy:
Excellent, this should be the default behavior for "reenroll", not sure why it is missed so long time.

cmd/fabric-ca-client/command/main_test.go
```diff
@@ -2531,7 +2531,11 @@ func registerAndRevokeUsers(t *testing.T, admin *lib.Identity, num int) []*big.I
 			// Reenroll the user, this should create a new certificate, so this
 			// user will have two valid certificates, but we will revoke one
 			// of her certificate only
-			_, err := enrollResp.Identity.Reenroll(&api.ReenrollmentRequest{})
+			_, err := enrollResp.Identity.Reenroll(&api.ReenrollmentRequest{
+				CSR: &api.CSRInfo{
+					KeyRequest: &api.KeyRequest{},
```
sykesm, line 2536: This should be optional, right? So why did the test need to change?

It's likely I'm missing something.

lib/client.go
```diff
@@ -191,7 +194,7 @@ func (c *Client) GetCAInfo(req *api.GetCAInfoRequest) (*GetCAInfoResponse, error
 }
 
 // GenCSR generates a CSR (Certificate Signing Request)
-func (c *Client) GenCSR(req *api.CSRInfo, id string) ([]byte, bccsp.Key, error) {
+func (c *Client) GenCSR(req *api.CSRInfo, id string, k bccsp.Key) ([]byte, bccsp.Key, error) {
```
sykesm, line 197: Backwards compatibility isn't necessarily a strong requirement but this is the client API. Do we need (or really want) to make this change to add an argument and overload the semantics based on the presence of the key?

lib/identity.go
```diff
@@ -130,7 +131,15 @@ func (i *Identity) RegisterAndEnroll(req *api.RegistrationRequest) (*Identity, e
...
-	csrPEM, key, err := i.client.GenCSR(req.CSR, i.GetName())
+	var key bccsp.Key
+	if req.CSR.KeyRequest.ReuseKey {
+		val, err := i.GetX509Credential().Val()
+		if err != nil {
+			return nil, err
+		}
+		key = val.(*x509.Signer).Key()
+	}
+	csrPEM, key, err := i.client.GenCSR(req.CSR, i.GetName(), key)
```
sykesm, line 142: Here is the meat of the change. Glancing at the code, I see `Reenroll` checking a nested struct in `CSRInfo` (no nil check - which goes to the comment I made in main_test) and using that information to extract a key based on `ReuseKey`. This structure seems to be motivating the change to `GenCSR` (breaking client API) to pass down the key.

My guess is that this PR would have a much smaller surface area and much smaller impact if a new interface function was introduced on the client to handle this re-enrollment flow with an existing key.

internal/pkg/api/client.go
```diff
@@ -86,6 +86,9 @@ type ReenrollmentRequest struct {
 	// AttrReqs are requests for attributes to add to the certificate.
 	// Each attribute is added only if the requestor owns the attribute.
 	AttrReqs []*AttributeRequest `json:"attr_reqs,omitempty"`
+	// ReuseKey specifies whether to reuse the existing private key
+	// when reenrolling an identity.
+	ReuseKey bool `json:"reusekey"`
```
sykesm, line 91: I'm very confused why this field exists at this level and also within the CSRInfo. It looks like the implementation of `Reenroll` does not look at this but only the nested value.

Can you explain the intent?

sykesm (changes requested):
I have a couple of questions about the PR - in particular about breaking the `lib/client` API and the duplication of the `ReuseKey` field across `api.ReenrollmentRequest` and `api.CSRInfo#KeyRequest`.

Thanks.

internal/pkg/api/client.go
wlahti: Good catch. This is leftover from my initial attempt. I'll remove it.

cmd/fabric-ca-client/command/main_test.go
wlahti: Fixed

lib/client.go
wlahti: Fixed

lib/identity.go
wlahti: Done

internal/pkg/api/client.go
wlahti: Done

sykesm (approved):

PR #196, opened 2020-10-20
Fix a typo in users-guide

naonishijima:
#### Type of change
- Documentation update

#### Description
affiliationchange should be affiliationchanged.


naonishijima:
RevocationReasonCodes map has "affiliationchanged" value, but the documents write "affiliationchange".

https://github.com/hyperledger/fabric-ca/blob/41d4b3f1ccefe03ba961d9f1fbd2cbd13bb995e0/internal/pkg/util/util.go#L70

saad-karim (approved):

PR #200, opened 2020-11-17
Move to Go 1.15.5

mastersingh24:
#### Type of change

Improvement (currency)

#### Description

Update to Go 1.15.5

### DIFF
--- a/lib/ca_test.go
+++ b/lib/ca_test.go
@@ -217,7 +217,7 @@ func testValidMatchingKeys(cert *x509.Certificate, t *testing.T) {
 		t.Error("Should have failed, public key and private key do not match")
 	}
 
-	err = validateMatchingKeys(cert, string(0))
+	err = validateMatchingKeys(cert, "/does-not-exist")
 	t.Log("validateMatchingKeys Error: ", err)
 	if err == nil {
 		t.Error("Should have failed to read bad file")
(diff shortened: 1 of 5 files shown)


btl5037:
Missed the Makefile: https://github.com/hyperledger/fabric-ca/blob/master/Makefile#L57

mastersingh24:
> Missed the Makefile: https://github.com/hyperledger/fabric-ca/blob/master/Makefile#L57

👎

lib/ca_test.go
```diff
@@ -217,7 +218,7 @@ func testValidMatchingKeys(cert *x509.Certificate, t *testing.T) {
 		t.Error("Should have failed, public key and private key do not match")
 	}
 
-	err = validateMatchingKeys(cert, string(0))
+	err = validateMatchingKeys(cert, strconv.Itoa(0))
```
sykesm, line 221: Just a comment: Using Itoa to render the number `0` as the string `"0"` could be done by simply replacing it with the literal `"0"`.

It also looks like most of these changes are related to trying to read or create files in bad paths.

mastersingh24: >Just a comment: Using Itoa to render the number 0 as the string "0" could be done by simply replacing it with the literal "0"

True enough

>It also looks like most of these changes are related to trying to read or create files in bad paths. I can't tell if the author was initially trying to introduce null terminators into the strings (rune(0) or "\x00") or just come up with a bogus value.

seems like both depending on the test

mastersingh24: I went with bogus strings to make the intention clearer.

sykesm (approved):
lgtm

PR #203, opened 2020-11-26
Certificate NotBefore date can not be before CA Cert

darekzbik:
#### Type of change

- Bug fix

#### Description

NotBefore data in certificate issued by fabric-ca can not be before CA cert used by this CA server.

#### Additional details

* unit test is added as a part of PR
#### Related issues

[FABC-832](https://jira.hyperledger.org/browse/FABC-832)

### DIFF
--- a/lib/serverenroll_test.go
+++ b/lib/serverenroll_test.go
@@ -7,13 +7,14 @@ SPDX-License-Identifier: Apache-2.0
 package lib
 
 import (
-	"os"
-	"testing"
-
+	"github.com/cloudflare/cfssl/signer"
 	"github.com/hyperledger/fabric-ca/internal/pkg/api"
 	"github.com/hyperledger/fabric-ca/internal/pkg/util"
 	dbuser "github.com/hyperledger/fabric-ca/lib/server/user"
 	"github.com/stretchr/testify/assert"
+	"os"
+	"testing"
+	"time"
 )
 
 func TestStateUpdate(t *testing.T) {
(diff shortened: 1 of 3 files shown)


lib/serverenroll_test.go
```diff
@@ -170,3 +171,53 @@ func TestPasswordLimit(t *testing.T) {
...
+	// verify response
+	certBytes, err := util.B64Decode(result.Cert)
+	assert.NoError(t, err, "Failed to convert certificate")
+	userCert, err := util.GetX509CertificateFromPEM(certBytes)
+	assert.NoError(t, err, "Failed to extract certificate from enroll response")
+
+	assert.False(t, userCert.NotBefore.Before(caCert.NotBefore),
+		"user certificate NotBefore %v is before CA cert NotBefore %v", userCert.NotBefore, caCert.NotBefore)
+	assert.False(t, userCert.NotAfter.After(caCert.NotAfter),
+		"user certificate NotAfter %v is after CA cert NotAfter %v", userCert.NotAfter, caCert.NotAfter)
```
mastersingh24, line 222: @darekzbik 

Thanks for taking this on.  Nothing wrong with what you've done, but a couple of suggestions:

While these checks are fine, shouldn't you actually check that NotBefore and NotAfter are equal to the CA cert values as well?

And should there also be a test where NotBefore and NotAfter from the enroll request are in the valid range?

mastersingh24 (approved):
I'll add the additional tests

lib/serverenroll_test.go
darekzbik: 1.  I did not check equality.  I was focused on checking if created certificate dates are in range of CA cert validity dates. Checking equality is fine as long as we assume that CA server is creating certificates with validity period as wide as possible. EQ test will fail when CA server will be changed and certs created by CA will be time limited (let say no longer than 360 days from now).

1. In this PR I touch only ‘NotBefore’ field in the cert. Logic of ‘userCert.NotAfter’ was previously in the code, probably not tested in the same way. I was focused on my change only. I’ll odd some more check in the same test.

In fact I'm surprised that CA server is issuing certs with past dates when someone ask for it. I don't like such behaviour but such change has potentially bigger impact.

darekzbik: @mastersingh24 plese see this https://github.com/hyperledger/fabric-ca/pull/206

PR #206, opened 2020-12-28
Small test improvement/refactor

darekzbik:
#### Type of change

- Test update

#### Description

Small test improvement/refactor related to conversation in previous PR https://github.com/hyperledger/fabric-ca/pull/203#discussion_r541558458 

#### Additional details

In PR #203  I modified code to code to ensure that created certificate is not before cert used by CA server.  During test
https://github.com/hyperledger/fabric-ca/blob/9ea68369cb61980c00a2a3518a8200c9bb54f70f/lib/serverenroll_test.go#L175 certificate request with validity range wider than CA cert is created. In this way I'm able to test **notBefore**  and expected **notBefore** in a new cert is is equal to **notBefore** in the CA cert.  

Unfortunately it is not so easy in case of **notAfter**. In the test I'm requesting for a validity time after CA cert **notAfter** https://github.com/hyperledger/fabric-ca/blob/9ea68369cb61980c00a2a3518a8200c9bb54f70f/lib/serverenroll_test.go#L197 . In the code **notAfter** time in certificate request seems to be  properly adjusted here: https://github.com/hyperledger/fabric-ca/blob/9ea68369cb61980c00a2a3518a8200c9bb54f70f/lib/serverenroll.go#L125 (this code was created long time ago). Unfortunately this adjustments is does not help too much because **notAfter** in the certificate is replaced a couple of lines earlier 
https://github.com/hyperledger/fabric-ca/blob/9ea68369cb61980c00a2a3518a8200c9bb54f70f/lib/serverenroll.go#L115

So in fact you can request for a certificate valid for 10minutes and CA is always generous and will create cert valid for 1year (profile.Expiry).

So this PR is just a little cleanup of tests. Comments and suggestions are welcome.

### DIFF
--- a/lib/serverenroll_test.go
+++ b/lib/serverenroll_test.go
@@ -216,8 +216,20 @@ func TestCertificateExpiration(t *testing.T) {
 	userCert, err := util.GetX509CertificateFromPEM(certBytes)
 	assert.NoError(t, err, "Failed to extract certificate from enroll response")
 
-	assert.False(t, userCert.NotBefore.Before(caCert.NotBefore),
-		"user certificate NotBefore %v is before CA cert NotBefore %v", userCert.NotBefore, caCert.NotBefore)
-	assert.False(t, userCert.NotAfter.After(caCert.NotAfter),
-		"user certificate NotAfter %v is after CA cert NotAfter %v", userCert.NotAfter, caCert.NotAfter)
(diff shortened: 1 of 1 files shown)


mastersingh24 (approved):
Thanks for the updated tests

PR #208, opened 2021-01-05
[FABC-909] Check If DB Exists Prior to Creating

lindluni:
Prior to creating the CA database, first determine if it exists in MySQL and Postgres databases. This is helpful when the DB user does not have permission to create the DB and the DB was created by an Admin out of band.

### DIFF
--- a/lib/server/db/mysql/mysql.go
+++ b/lib/server/db/mysql/mysql.go
@@ -8,6 +8,8 @@ package mysql
 
 import (
 	"context"
+	"database/sql"
+	"fmt"
 	"regexp"
 	"strings"
 
@@ -93,6 +95,18 @@ func (m *Mysql) PingContext(ctx context.Context) error {
 	return nil
 }
 
+// exists determines if the database has already been created
+func exists(sqlxDB db.FabricCADB, dbName string) (bool, error) {
+	log.Debugf("Checking if MySQL CA database '%s' exists", dbName)
+	var exists bool
+	query := fmt.Sprintf("SELECT true as 'exists' FROM INFORMATION_SCHEMA.SCHEMATA WHERE SCHEMA_NAME = '%s'", dbName)
+	err := sqlxDB.Get("CheckIfDatabaseExists", &exists, query)
+	if err != nil && err != sql.ErrNoRows {
+		return false, errors.Wrapf(err, "Failed to check if MySQL CA database '%s' exists", dbName)
+	}
+	return exists, nil
+}
+
 // Create creates database and tables
 func (m *Mysql) Create() (*db.DB, error) {
 	db, err := m.CreateDatabase()
@@ -108,19 +122,16 @@ func (m *Mysql) Create() (*db.DB, error) {
 
 // CreateDatabase creates database
 func (m *Mysql) CreateDatabase() (*db.DB, error) {
-	datasource := m.datasource
-	dbName := m.dbName
 	err := m.createDatabase()
 	if err != nil {
 		return nil, errors.Wrap(err, "Failed to create MySQL database")
 	}
 
-	log.Debugf("Connecting to database '%s', using connection string: '%s'", dbName, util.MaskDBCred(datasource))
-	sqlxdb, err := sqlx.Open("mysql", datasource)
+	log.Debugf("Connecting to database '%s', using connection string: '%s'", m.dbName, util.MaskDBCred(m.datasource))
+	sqlxdb, err := sqlx.Open("mysql", m.datasource)
 	if err != nil {
-		return nil, errors.Wrapf(err, "Failed to open database (%s) in MySQL server", dbName)
+		return nil, errors.Wrapf(err, "Failed to open database '%s' in MySQL server", m.dbName)
 	}
-
 	m.SqlxDB = db.New(sqlxdb, m.CAName, m.Metrics)
 
 	return m.SqlxDB.(*db.DB), nil
@@ -136,14 +147,17 @@ func (m *Mysql) CreateTables() error {
 }
 
 func (m *Mysql) createDatabase() error {
-	dbName := m.dbName
-	log.Debugf("Creating MySQL Database (%s) if it does not exist...", dbName)
-
-	_, err := m.SqlxDB.Exec("CreateDatabase", "CREATE DATABASE IF NOT EXISTS "+dbName)
+	exists, err := exists(m.SqlxDB, m.dbName)
 	if err != nil {
-		return errors.Wrap(err, "Failed to execute create database query")
+		return err
 	}
-
+	if !exists {
+		log.Debugf("Creating MySQL Database '%s'", m.dbName)
+		_, err := m.SqlxDB.Exec("CreateDatabase", "CREATE DATABASE "+m.dbName)
+		if err != nil {
+			return errors.Wrap(err, "Failed to execute create database query")
+		}
+	}
 	return nil
 }
 
--- a/lib/server/db/mysql/mysql_test.go
+++ b/lib/server/db/mysql/mysql_test.go
@@ -4,14 +4,13 @@ Copyright IBM Corp. All Rights Reserved.
 SPDX-License-Identifier: Apache-2.0
 */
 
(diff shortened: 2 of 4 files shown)


lindluni:
Per my comments in #156 pushing it up into the `ca.go` was much more intrusive than I expected. So went with the local private functions and smaller surface area.

lib/server/db/postgres/postgres_test.go
```diff
@@ -4,13 +4,12 @@ Copyright IBM Corp. All Rights Reserved.
 SPDX-License-Identifier: Apache-2.0
 */
 
-package postgres_test
+package postgres
```
sykesm, line 7: If we're using a mock, why do we need to access the unexported `exists` function? We can verify the interaction and parameters by calling `CreateDatabase`.

lib/server/db/mysql/mysql.go
```diff
@@ -93,6 +95,18 @@ func (m *Mysql) PingContext(ctx context.Context) error {
...
+func exists(sqlxDB db.FabricCADB, dbName string) (bool, error) {
+	log.Debugf("Checking if MySQL CA Database (%s) exists", dbName)
+	var exists bool
+	query := fmt.Sprintf("SELECT true as 'exists' FROM INFORMATION_SCHEMA.SCHEMATA WHERE SCHEMA_NAME = '%s'", dbName)
+	err := sqlxDB.Get("CheckIfDatabaseExists", &exists, query)
+	if err != nil && err != sql.ErrNoRows {
+		return false, errors.Wrapf(err, "Failed to check if MySQL database '%s' exists", dbName)
+	}
+	return exists, nil
+}
```
sykesm, line 108: Why is this needed for MySQL? The CREATE DATABASE statement supports the IF NOT EXISTS predicate for MySQL and it's already used in the createDatabase function below.

I thought the reported problem was for postgres where the IF NOT EXISTS predicate is not supported?

lib/server/db/postgres/postgres_test.go
```diff
@@ -133,6 +132,10 @@ var _ = Describe("Postgres", func() {
 			_, err := db.CreateDatabase()
 			Expect(err).To(HaveOccurred())
 			Expect(err.Error()).Should(ContainSubstring("Failed to create Postgres database: Failed to execute create database query: error creating database"))
+
+			exists, err := exists(db.SqlxDB, db.dbName)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(exists).To(BeFalse())
```
sykesm, line 138: The existence query should be a separate context and assertion. Something like `When("the database already exists") ... It("does not attempt to create the database")...`.

lib/server/db/postgres/postgres.go
```diff
@@ -105,6 +106,18 @@ func (p *Postgres) PingContext(ctx context.Context) error {
 	return nil
 }
 
+// Exists determines if the database has already been created
```
sykesm, line 109: Having this up here doesn't quite align with the debug messages issued by `createDatabase` below. Perhaps the call to should be moved into `createDatabase`?

lib/server/db/postgres/postgres.go
```diff
@@ -120,12 +133,18 @@ func (p *Postgres) Create() (*db.DB, error) {
...
 func (p *Postgres) CreateDatabase() (*db.DB, error) {
-	dbName := p.dbName
-	err := p.createDatabase()
+	exists, err := exists(p.SqlxDB, p.dbName)
 	if err != nil {
-		return nil, errors.Wrap(err, "Failed to create Postgres database")
+		return nil, err
+	}
+	if !exists {
+		err = p.createDatabase()
```
sykesm, line 141: In `createDatabase` we still have the following block of code.

```
	if err != nil {
		if !strings.Contains(err.Error(), fmt.Sprintf("database \"%s\" already exists", dbName)) {
			return errors.Wrap(err, "Failed to execute create database query")
		}
	}
```

Shouldn't that be killed now?

sykesm (changes requested):
I don't think that the MySQL path needs to change but maybe I'm missing something.
For postgres, checking the catalog for the database seems like the correct thing to do but there's some fallout that should be fixed up like the debug messages and the error message checks related to creation.

I also think the tests should focus on the behavior instead of simply exercising the `exists` function. When we're driving `CreateDatabase`, the assertions should describe what happens when a database already exists.

lib/server/db/mysql/mysql.go
lindluni: I tested this. If the user has the `CREATE` permission on the database which grants `Database, Table and Column` permission, `IF NOT EXISTS` works. But if the user only has `Insert` permission they have  `Table and Column` permission and `IF NOT EXISTS` returns `Access denied for user 'btl5037'@'%' to database 'test'`. With create they can modify the existing database. I imagine if you were locking down permissions you would provide `Insert` and not `Create` permission. But I could be wrong, thought?

lib/server/db/postgres/postgres.go
lindluni: Agreed

lib/server/db/postgres/postgres.go
lindluni: Missed this, thank you

lindluni:
Just need to modify the `exists` test to have `mock.Get` return true, other comments addressed

lindluni:
I think I addressed everything, but I'm not sure if I did the typecasting of the interface on the mock correct to make sure the value was returned and not lost. Maybe there was a more idiomatic way you could share.

My comments on MySQL are still up for debate.

Thanks Matt.

lib/server/db/postgres/postgres_test.go
```diff
@@ -127,19 +126,40 @@ var _ = Describe("Postgres", func() {
...
+			})
+		})
+
+		When("the database already exists", func() {
+			It("does not attempt to create the database", func() {
+				mockDB.GetCalls(func(s string, i interface{}, s2 string, i2 ...interface{}) error {
+					exists := i.(*bool)
+					*exists = true
+					i = exists
+					return nil
```
lindluni, line 155: Not sure this is the best way to do it, but this is what I came up with to make sure the value assigned to the interface sticks.

lib/server/db/mysql/mysql_test.go
```diff
@@ -82,19 +81,40 @@ var _ = Describe("Mysql", func() {
...
+		})
+
+		When("the database already exists", func() {
+			It("does not attempt to create the database", func() {
+				mockDB.GetCalls(func(s string, i interface{}, s2 string, i2 ...interface{}) error {
+					exists := i.(*bool)
+					*exists = true
+					i = exists
+					return nil
+				})
```
sykesm, line 111:
```suggestion
				mockDB.GetStub = func(s string, i interface{}, s2 string, i2 ...interface{}) error {
					*(i.(*bool)) = true
					return nil
				}
```

lib/server/db/postgres/postgres_test.go
```diff
@@ -127,19 +126,40 @@ var _ = Describe("Postgres", func() {
...
+				})
```
sykesm, line 156:
```suggestion
				mockDB.GetStub = func(s string, i interface{}, s2 string, i2 ...interface{}) error {
					*(i.(*bool)) = true
					return nil
				}
```

lib/server/db/mysql/mysql_test.go
```diff
@@ -4,14 +4,13 @@ Copyright IBM Corp. All Rights Reserved.
 SPDX-License-Identifier: Apache-2.0
 */
 
-package mysql_test
+package mysql
```
sykesm, line 7: This should go back to the `mysql_test` package, right?

lib/server/db/postgres/postgres_test.go
sykesm, line 7: This should go back to the `postgres_test` package, right?

sykesm (commented):
Looks okay to me but I think the test changes could be more targeted and simpler.

sykesm (approved):

PR #210, opened 2021-01-12
[FABC-905] Add passfile to server config file template

stephyee:
#### Type of change

- Bug fix

#### Description

Add server config option passfile. This config will take precedence over password and will not be stored in the generated yaml. 

#### Related issues
[FABC-905](https://jira.hyperledger.org/browse/FABC-905)

### DIFF
--- a/cmd/fabric-ca-server/config.go
+++ b/cmd/fabric-ca-server/config.go
@@ -10,6 +10,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	"github.com/cloudflare/cfssl/log"
@@ -160,6 +161,7 @@ registry:
   identities:
      - name: <<<ADMIN>>>
        pass: <<<ADMINPW>>>
+       passFile: <<<ADMINPASSFILE>>>
        type: client
        affiliation: ""
        attrs:
@@ -619,7 +621,7 @@ func (s *ServerCmd) configInit() (err error) {
 }
 
 func (s *ServerCmd) createDefaultConfigFile() error {
-	var user, pass string
+	var user, pass, upFile string
 	// If LDAP is enabled, authentication of enrollment requests are performed
 	// by using LDAP authentication; therefore, no bootstrap username and password
 	// are required.
@@ -631,6 +633,7 @@ func (s *ServerCmd) createDefaultConfigFile() error {
 		// bootstrap administrator.  Other identities can be dynamically registered.
 		// Create the default config, but only if they provided this bootstrap
 		// username and password.
+		upFile = s.myViper.GetString("bootfile")
 		up := s.myViper.GetString("boot")
 		if up == "" {
 			return errors.New("The '-b user:pass' option is required")
@@ -644,11 +647,12 @@ func (s *ServerCmd) createDefaultConfigFile() error {
 		}
 		user = ups[0]
 		pass = ups[1]
+
 		if len(user) >= 1024 {
 			return errors.Errorf("The identity name must be less than 1024 characters: '%s'", user)
 		}
-		if len(pass) == 0 {
-			return errors.New("An empty password in the '-b user:pass' option is not permitted")
+		if len(pass) == 0 && upFile == "" {
+			return errors.New("An empty password in the '-b user:pass' and '-f passfile' option is not permitted")
 		}
 	}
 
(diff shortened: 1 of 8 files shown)


cmd/fabric-ca-server/config.go
```diff
@@ -12,6 +12,8 @@ import (
 	"path/filepath"
 	"strings"
 
+	"regexp"
```
wlahti, line 15: Super nit: this should be grouped with the above imports

cmd/fabric-ca-server/main_test.go
```diff
@@ -118,20 +107,48 @@ func TestErrors(t *testing.T) {
...
+		testName string
+		cmd      []string
+	}{
+		{
+			testName: "When identity has user and pass",
+			cmd:      []string{cmdName, "init", "-b", "admin:adminpw", "--registry.maxenrollments", "1", "-H", os.TempDir()},
+		},
+		{
+			testName: "When identity has user and passfile",
+			cmd:      []string{cmdName, "init", "-f", "pwFile", "--registry.maxenrollments", "1", "-H", os.TempDir()},
```
wlahti, line 133: Where does the user come for this test command? I'm not super familiar with fabric-ca but I was expecting to see the -b option here for the username.

lib/ca_test.go
```diff
@@ -351,6 +352,98 @@ func TestCAgetUserAttrValue(t *testing.T) {
 	CAclean(ca, t)
 }
 
+func TestCAAddIdentity(t *testing.T) {
+
```
wlahti, line 356: Super nit: empty line

lib/ca_test.go
```diff
@@ -351,6 +352,98 @@ func TestCAgetUserAttrValue(t *testing.T) {
...
+			id: &CAConfigIdentity{
+				Name:     "admin3",
+				Pass:     "adminpw",
+				PassFile: filepath.Join(os.TempDir(), "adminpwfile"),
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
```
wlahti, line 386: Nit: this isn't necessary unless the tests are run with `t.Parallel()`, right?

cmd/fabric-ca-server/servercmd.go
```diff
@@ -154,6 +154,8 @@ func (s *ServerCmd) registerFlags() {
 	pflags.StringVarP(&s.homeDirectory, "home", "H", "", fmt.Sprintf("Server's home directory (default \"%s\")", filepath.Dir(cfg)))
 	util.FlagString(s.myViper, pflags, "boot", "b", "",
 		"The user:pass for bootstrap admin which is required to build default config file")
+	util.FlagString(s.myViper, pflags, "bootfile", "f", "",
```
wlahti, line 157: Usage question here: when I try to specify a username and passfile, it complains because I haven't specified a password after a colon separator:
`fabric-ca-server init -b admin -f pass.file`
`Error: Failed to create default configuration file: The value 'admin' on the command line is missing a colon separator`

While it isn't too difficult to add the `: ` after `admin`, it feels like we shouldn't require that when `-f` is used. Any reason that shouldn't be the case?

cmd/fabric-ca-server/main_test.go
stephyee: ah, this was oversight in the test definition. The test passes because there are no checks around the user existence so the value for user remains `<<<ADMIN>>>`.

cmd/fabric-ca-server/config.go
stephyee: Courtesy of GoImports. 🧹

lib/ca_test.go
stephyee: 🧹

lib/ca_test.go
stephyee: Correct, initially I tried to run these tests in parallel but this caused an issue because of the shared artifacts. 🧹

cmd/fabric-ca-server/servercmd.go
stephyee: As we discussed, this was a design choice (trying to change as little as possible). The `init` and `start` commands still require both `-b admin:pass` with the `-f pass.file` flag specified. The _pass_ from `-b user:pass` will be used as a backup when a password cannot be read from _pass.file_ when starting server. 

Another option as you suggested is to allow only `-b admin` when `-f pass.file` is provided then
1.  Allow the server to error on startup if the _pass.file_ cannot be read. 
2. Error when creating the default config file if the _pass.file_ doesn't exist, cannot be read, etc.

wlahti: @sykesm Any thoughts / opinions here?

ryjones:
@hyperledger/fabric-core-maintainers @hyperledger/fabric-ca-maintainers I attempted a rebase to resolve the conflicts, but I don't know the code well enough to make a choice. Could someone please rebase this for merging?

PR #214, opened 2021-02-22
Delete undefined variable in lib/server_test.go

naonishijima:
#### Type of change
- Bug fix

#### Description
When execute go test server_test.go in local, it will be failed due to undefined "serversDir" variable.
I supposed "serversDir" variable is no longer use in server_test.go.


sykesm:
@naonishijima I'm not really sure about this PR. I don't see any failures when I run this locally and the var is defined at the package level (yuck) in `client_test.go`.

Can you please include the output of the failure you're seeing as well as the platform you're running on?

naonishijima:
Sorry, I misunderstand this. I tried to test again. It was OK.

PR #215, opened 2021-02-22
Delete calog.SetLogLevel() in server.go

naonishijima:
#### Type of change

#### Description
calog.SetLogLevel() was called twice in server.go and config.go. This PR fixes to call it only once.


sykesm:
Thanks for this PR but there are a number of config related paths that have weird side effects in both Fabric and Fabric-CA. If this is not actually causing a regression, I'd prefer to leave it as is.

Is the (conditionally) duplicate call causing an issue?

sykesm:
Please reopen if this is more than a minor cleanup. Thanks.

PR #216, opened 2021-02-22
Change test function to a suitable name

naonishijima:
#### Type of change

#### Description
There is a similar name in serveraffiliations_test.go and the name is not appropriate. This PR changes the name to TestServerAffiliation which like other function name rules(TestServerxxx) in the same file.

### DIFF
--- a/lib/server_whitebox_test.go
+++ b/lib/server_whitebox_test.go
@@ -29,12 +29,12 @@ const (
 	affiliationName = "org1"
 )
 
-// TestGetAffliation checks if there is one record for the
+// TestServerGetAllAffiliations checks if there is one record for the
 // affilition 'org1' in the database after starting the server
 // two times. This test is to make sure server does not create
 // duplicate affiliations in the database every time it is
 // started.
-func TestGetAffliation(t *testing.T) {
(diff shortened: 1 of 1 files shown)


lib/server_whitebox_test.go
```diff
@@ -29,12 +29,12 @@ const (
...
 )
 
-// TestGetAffliation checks if there is one record for the
+// TestServerAffiliation checks if there is one record for the
 // affilition 'org1' in the database after starting the server
 // two times. This test is to make sure server does not create
 // duplicate affiliations in the database every time it is
 // started.
-func TestGetAffliation(t *testing.T) {
+func TestServerAffilication(t *testing.T) {
```
sykesm, line 37: If we're renaming, can we include the full method name?

`TestServerGetAllAffiliations` ?

sykesm (changes requested):

lib/server_whitebox_test.go
naonishijima: Thank you for your review. I got it. I will do that.

sykesm (approved):
Thank you.

PR #231, opened 2021-03-29
[FABC-930] Overwrite CN field to allow CSR modification from sdk

vieiramanoel:
Signed-off-by: Vieira Neto <vieiranetoc@gmail.com>

Type
######
Lib CSR creation

- Bug fix

#### Description

When coding from fab-sdk-go the function `msp.WithCSR(&msp.CSRInfo{CN: commonName})` has no effect on generated certificate as the lib's function `newCertificateRequest` ignores this field from request. 
My suggestion is to overwrite the CN fields when it exists.

#### Additional details

I'm aiming to write some tests for this change but i couldn't find the correct test to be modified/where to add. Hope someone can help me with this!

#### Related issues

https://jira.hyperledger.org/browse/FABG-1037

#### Release Note
 
- CN field can now be set in lib's api

### DIFF
--- a/lib/client.go
+++ b/lib/client.go
@@ -555,6 +555,10 @@ func (c *Client) newCertificateRequest(req *api.CSRInfo, id string) *csr.Certifi
 		}
 		cr.KeyRequest = newCfsslBasicKeyRequest(keyRequest)
 
+		if req.CN != "" {
+			cr.CN = req.CN
+		}
+
 		return cr
 	}
 


mastersingh24:
@vieiramanoel  - this is actually by design.  We don't allow people to create arbitrary CNs in the via the Fabric CA APIs.  As you can see from the function signature, we always use the enrollment ID as the CN.  This was to prevent users from creating conflicting CNs as well as to prevent users from masquerading as other users when code uses the CN to identity users in chaincode or other resources.

mastersingh24 (changes requested):
Putting a block here to prevent merging.  See my earlier comment.

PR #238, opened 2021-06-22
[FABC-931] Re-enroll with expired certs 

mbwhite:
#### Type of change

- Documentation update

#### Description

Resolves to https://jira.hyperledger.org/browse/FABC-931.
New config option to allow for certificate expiry to be ignored in the case of a re-enroll of an id. But only in the re-enroll case.

### DIFF
--- a/docs/source/servercli.rst
+++ b/docs/source/servercli.rst
@@ -20,6 +20,7 @@ Fabric-CA Server's CLI
       -b, --boot string                               The user:pass for bootstrap admin which is required to build default config file
           --ca.certfile string                        PEM-encoded CA certificate file (default "ca-cert.pem")
           --ca.chainfile string                       PEM-encoded CA chain file (default "ca-chain.pem")
+          --ca.ignorecertexpiry                       Ignore Certificate Expirty for re-enroll
           --ca.keyfile string                         PEM-encoded CA key file
       -n, --ca.name string                            Certificate Authority name
           --cacount int                               Number of non-default CA instances
--- a/docs/source/servercli.rst
+++ b/docs/source/servercli.rst
@@ -20,9 +20,9 @@ Fabric-CA Server's CLI
       -b, --boot string                               The user:pass for bootstrap admin which is required to build default config file
           --ca.certfile string                        PEM-encoded CA certificate file (default "ca-cert.pem")
           --ca.chainfile string                       PEM-encoded CA chain file (default "ca-chain.pem")
-          --ca.ignorecertexpiry                       Ignore Certificate Expirty for re-enroll
           --ca.keyfile string                         PEM-encoded CA key file
       -n, --ca.name string                            Certificate Authority name
+          --ca.reenrollignorecertexpiry               Ignore Certificate Expirty for re-enroll
           --cacount int                               Number of non-default CA instances
           --cafiles strings                           A list of comma-separated CA configuration files
           --cfg.affiliations.allowremove              Enables removal of affiliations dynamically
--- a/docs/source/servercli.rst
+++ b/docs/source/servercli.rst
@@ -22,7 +22,7 @@ Fabric-CA Server's CLI
           --ca.chainfile string                       PEM-encoded CA chain file (default "ca-chain.pem")
           --ca.keyfile string                         PEM-encoded CA key file
       -n, --ca.name string                            Certificate Authority name
-          --ca.reenrollignorecertexpiry               Ignore Certificate Expirty for re-enroll
+          --ca.reenrollignorecertexpiry               Ignore Certificate Expiry for re-enroll
           --cacount int                               Number of non-default CA instances
           --cafiles strings                           A list of comma-separated CA configuration files
           --cfg.affiliations.allowremove              Enables removal of affiliations dynamically
--- a/lib/ca.go
+++ b/lib/ca.go
@@ -473,7 +473,7 @@ func (ca *CA) initConfig() (err error) {
 // Return nil if successful; otherwise, return an error.
 func (ca *CA) VerifyCertificate(cert *x509.Certificate, forceTime bool) error {
 
-	log.Debugf("Certicate Dates: NotAfter = %s\n Cert NotBefore = %s \n", cert.NotAfter.String(), cert.NotBefore.String())
+	log.Debugf("Certicate Dates: NotAfter = %s NotBefore = %s \n", cert.NotAfter.String(), cert.NotBefore.String())
 
 	opts, err := ca.getVerifyOptions()
 	if err != nil {
--- a/lib/ca.go
+++ b/lib/ca.go
@@ -471,6 +471,10 @@ func (ca *CA) initConfig() (err error) {
 
 // VerifyCertificate verifies that 'cert' was issued by this CA
 // Return nil if successful; otherwise, return an error.
+// 'forceTime' if false, certificate expiry times will be checked based
+// on the current time.
+// if true, it will force the time to be used to check for expiry to be 30 seconds
+// after the certificate start time.  (this is to support reenrollIgnoreCertExpiry)
 func (ca *CA) VerifyCertificate(cert *x509.Certificate, forceTime bool) error {
 
 	log.Debugf("Certicate Dates: NotAfter = %s NotBefore = %s \n", cert.NotAfter.String(), cert.NotBefore.String())
@@ -480,8 +484,8 @@ func (ca *CA) VerifyCertificate(cert *x509.Certificate, forceTime bool) error {
 		return errors.WithMessage(err, "Failed to get verify options")
 	}
 
-	// force time to be 30seconds after start to ensure expiry doesn't get flaged
-	// this is one of the checks that made on the certificate
+	// force check time to be 30 seconds after certificate start time to ensure expiry doesn't get flagged
+	// this is one of the checks that is made on the certificate in Verify()
 	if forceTime {
 		opts.CurrentTime = cert.NotBefore.Add(time.Duration(time.Second * 30))
 	}
--- a/lib/ca.go
+++ b/lib/ca.go
@@ -471,11 +471,21 @@ func (ca *CA) initConfig() (err error) {
 
 // VerifyCertificate verifies that 'cert' was issued by this CA
 // Return nil if successful; otherwise, return an error.
-func (ca *CA) VerifyCertificate(cert *x509.Certificate) error {
+func (ca *CA) VerifyCertificate(cert *x509.Certificate, forceTime bool) error {
+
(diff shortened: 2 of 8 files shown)


lib/ca.go
```diff
@@ -471,11 +471,21 @@ func (ca *CA) initConfig() (err error) {
 
 // VerifyCertificate verifies that 'cert' was issued by this CA
 // Return nil if successful; otherwise, return an error.
-func (ca *CA) VerifyCertificate(cert *x509.Certificate) error {
+func (ca *CA) VerifyCertificate(cert *x509.Certificate, forceTime bool) error {
```
denyeart, line 474:
```suggestion
func (ca *CA) VerifyCertificate(cert *x509.Certificate, ignoreCertExpiry bool) error {
```

I think the boolean name should tie back to the config option name.

lib/caconfig.go
```diff
@@ -49,6 +49,8 @@ ca:
   certfile: ca-cert.pem
   # The CA key file
   keyfile: ca-key.pem
+  # Ignore Certificate Expiration in the case of re-enroll
+  ignoreCertExpiry: false
```
denyeart, line 53: The precedent is all lowercase.

docs/source/servercli.rst
```diff
@@ -20,6 +20,7 @@ Fabric-CA Server's CLI
       -b, --boot string                               The user:pass for bootstrap admin which is required to build default config file
           --ca.certfile string                        PEM-encoded CA certificate file (default "ca-cert.pem")
           --ca.chainfile string                       PEM-encoded CA chain file (default "ca-chain.pem")
+          --ca.ignorecertexpiry                       Ignore Certificate Expirty for re-enroll
```
denyeart, line 23: Since it only applies to reenroll and not other functions, we should probably include that in the property name (at the risk of it being too long), e.g. `ignorereenrollcertexpiry`

lib/serverrequestcontext.go
```diff
@@ -188,11 +188,13 @@ func (ctx *serverRequestContextImpl) verifyX509Token(ca *CA, authHdr, method, ur
 	if err2 != nil {
 		return "", caerrors.NewAuthenticationErr(caerrors.ErrInvalidToken, "Invalid token in authorization header: %s", err2)
 	}
+
 	// Make sure the caller's cert was issued by this CA
-	err2 = ca.VerifyCertificate(cert)
+	err2 = ca.VerifyCertificate(cert, (ctx.endpoint.Path == "reenroll" && ctx.ca.Config.CA.IgnoreCertExpiry))
```
denyeart, line 193: While this works I find it difficult to read when too much is done in a single line.
How about setting a variable whose name matches the parameter name prior to the function call and then passing the variable, e.g.

```
ignoreCertExpiry := ctx.endpoint.Path == "reenroll" && ctx.ca.Config.CA.IgnoreReenrollCertExpiry
err2 = ca.VerifyCertificate(cert,ignoreCertExpiry)
```

lib/serverrequestcontext.go
```diff
@@ -202,10 +204,17 @@ func (ctx *serverRequestContextImpl) verifyX509Token(ca *CA, authHdr, method, ur
 	if !checked {
 		return "", caerrors.NewHTTPErr(401, caerrors.ErrCertRevokeCheckFailure, "Failed while checking for revocation")
 	}
+
 	if expired {
-		return "", caerrors.NewAuthenticationErr(caerrors.ErrCertExpired,
-			"The certificate in the authorization header is a revoked or expired certificate")
+		log.Debugf("Expired Certificate")
+		if ctx.endpoint.Path == "reenroll" && ctx.ca.Config.CA.IgnoreCertExpiry {
```
denyeart, line 210: ...and if we create the well-named variable above, we can re-use it here instead of re-doing the logic.

lib/serverrequestcontext.go
```diff
@@ -202,10 +204,17 @@ func (ctx *serverRequestContextImpl) verifyX509Token(ca *CA, authHdr, method, ur
...
+			log.Infof("Ignoring expired certificate for re-endroll operation")
```
denyeart, line 211:
```suggestion
			log.Infof("Ignoring expired certificate for re-enroll operation")
```

lib/caconfig.go
denyeart: I'm new at fabric-ca config... when do config options go into lib/caconfig.go versus in cmd/fabric-ca-server/config.go? I looked at some of the prior config option commits, and most of the config options go into cmd/fabric-ca-server/config.go, while just the highest level config options go in lib/caconfig.go.

denyeart (commented):
Need a test for the new config option... unfortunately I don't know enough about how fabric-ca integration test works to provide more specific guidance.

lib/ca.go
mbwhite: > Need a test for the new config option... unfortunately I don't know enough about how fabric-ca integration test works to provide more specific guidance.

Yes absoltuley.

lib/caconfig.go
mbwhite: What the difference is between the 2, I'm not really sure; there are no comments in the code;  the `cmd/fabric-ca-server/config.go` seems to have the more complete file, and the call stack of the other config seems to be around multicast?  Can move it.

mbwhite: In terms of naming... I've seen `clientAuthRequired`  admit that `ignorereenrollcertexpiry` for me is harder to read than `ignore ReenrollCertExpiry`

lib/serverrequestcontext.go
mbwhite: ah yes.. it is used twice... will pull it out into it's own variable

lib/ca.go
mbwhite: on the name, I'd debated on tieing this to the ignoreCertExpiry but then decided against it
- there are multiple functions that do certificate checks - they are given a cert and then say yes/no;  there is some overlap between the checks they make - expiy being an example
- the controlling function is `verifyX509Token` - by renaming the variable it is pushing knowledge of the config option down into a function that doesn't actually need to know this.
- the controlling function is request a certain style of validation based on what it needs - in this case it needs to have the expiry time be discounted.  

Admit that the multiple checks is a wasteful in processing - and I don't like having to put a defined time to force the expiry check to not be an issue. But changing that is beyond the scope or this pr.

lib/ca.go
```diff
@@ -471,11 +471,21 @@ func (ca *CA) initConfig() (err error) {
...
+
+	log.Debugf("Certicate Dates: NotAfter = %s NotBefore = %s \n", cert.NotAfter.String(), cert.NotBefore.String())
+
 	opts, err := ca.getVerifyOptions()
 	if err != nil {
 		return errors.WithMessage(err, "Failed to get verify options")
 	}
+
+	// force time to be 30seconds after start to ensure expiry doesn't get flaged
+	// this is one of the checks that made on the certificate
```
denyeart, line 484:
```suggestion
	// force check time to be 30 seconds after certificate start time to ensure expiry doesn't get flagged
	// this is one of the checks that is made on the certificate in Verify()
```

lib/ca.go
denyeart: ok, let's at least add "forceTime" to the function comment on line 472 and mention that it is used if reenrollIgnoreCertExpiry is set, so that people know when this is used.

lib/caconfig.go
```diff
@@ -116,10 +116,11 @@ type affiliationsOptions struct {
...
 type CAInfo struct {
-	Name      string `opt:"n" help:"Certificate Authority name"`
-	Keyfile   string `help:"PEM-encoded CA key file"`
-	Certfile  string `def:"ca-cert.pem" help:"PEM-encoded CA certificate file"`
-	Chainfile string `def:"ca-chain.pem" help:"PEM-encoded CA chain file"`
+	Name                     string `opt:"n" help:"Certificate Authority name"`
+	Keyfile                  string `help:"PEM-encoded CA key file"`
+	Certfile                 string `def:"ca-cert.pem" help:"PEM-encoded CA certificate file"`
+	Chainfile                string `def:"ca-chain.pem" help:"PEM-encoded CA chain file"`
+	ReenrollIgnoreCertExpiry bool   `def:"false" help:"Ignore Certificate Expirty for re-enroll"`
```
denyeart, line 123:
```suggestion
	ReenrollIgnoreCertExpiry bool   `def:"false" help:"Ignore Certificate Expiry for re-enroll"`
```

lib/serverrequestcontext.go
```diff
@@ -188,11 +188,14 @@ func (ctx *serverRequestContextImpl) verifyX509Token(ca *CA, authHdr, method, ur
 	if err2 != nil {
 		return "", caerrors.NewAuthenticationErr(caerrors.ErrInvalidToken, "Invalid token in authorization header: %s", err2)
 	}
+
 	// Make sure the caller's cert was issued by this CA
-	err2 = ca.VerifyCertificate(cert)
+	reenrollIgnoreCertExpiry := ctx.endpoint.Path == "reenroll" && ctx.ca.Config.CA.ReenrollIgnoreCertExpiry
```
denyeart, line 193: The existing comment right above this line doesn't go with this line so it is a bit confusing.
I'd keep the existing comment with the ca.VerifyCertificate() line, and add a separate comment for this new line explaining that if it is true then we'll forceTime in the VerifyCertificate() check.

test/integration/certexpiry/certexpiry_test.go
```diff
@@ -0,0 +1,165 @@
...
+
+	// Enroll and then reenroll to check
+	err = command.RunMain([]string{cmdName, "enroll", "-u", userServiceEnrollURL, "-d"})
+	util.FatalError(t, err, "Failed to enroll user1")
+
+	err = command.RunMain([]string{cmdName, "reenroll", "-u", userServiceEnrollURL, "-d", "--csr.keyrequest.reusekey"})
+	util.FatalError(t, err, "Failed to reenroll user1")
+
+	log.Infof("Tested re-enroll of id, waiting for cert to expiry before testing re-enroll\n")
+	time.Sleep(2 * time.Minute)
```
denyeart, line 93: could we make the expiry in seconds rather than a minute? I hate to sleep for two whole minutes in a test... they add up...

denyeart (commented):
Looks pretty good now.
I've actually tested the PR end-to-end myself as a user to ensure it works as expected for expired TLS certs... it does!

test/integration/certexpiry/certexpiry_test.go
mbwhite: agree with you; didn't find sub minute timeouts reliable in testing however..

denyeart (approved):

PR #284, opened 2022-04-06
test: use `T.TempDir` to create temporary test directory

Juneezee:
#### Type of change

- Test update

#### Description

A testing cleanup. 

This pull request replaces `ioutil.TempDir` with `t.TempDir`. We can use the `T.TempDir` function from the `testing` package to create temporary directory. The directory created by `T.TempDir` is automatically removed when the test and all its subtests complete. 

Reference: https://pkg.go.dev/testing#T.TempDir


denyeart (approved):
Thank you!

PR #288, opened 2022-04-14
Make the debug flag visible in --help and docs

jkopczyn:
#### Type of change

- Documentation update

#### Description

The `-d` flag is used by default in test/demo materials such as hyperledger/fabric-samples, so having the the flag be hidden from the documentation causes confusion. Remove 'hide'; add a warning for why users might not want to use it, since those drawbacks were probably the primary reason it was hidden before.


jkopczyn:
No test failures seem to be a result of anything in this PR. They fail on main.

denyeart:
Close and re-open to trigger CI.

denyeart:
Finally figured out the problem... since the --debug flag wasn't added in alphabetical order, the generated doc didn't match the doc provided in this PR. The change has been submitted and merged in #371.

PR #290, opened 2022-05-15
escape special charecters that can be set through the command line ar…

RohanShrothrium:
…guments

Signed-off-by: Rohan Shrothrium <rohan.shrothrium@intellecteu.com>

#### Type of change

- Bug fix

#### Description

When we initialise a fabric ca server with an admin and adminpw, if they have special characters the ca behaves not as expected.
These special characters can be in anything that can be passed through the command line arguments.
This PR is meant to escape these special characters in the `yml` file.

### DIFF
--- a/cmd/fabric-ca-server/config.go
+++ b/cmd/fabric-ca-server/config.go
@@ -68,7 +68,7 @@ const (
 #############################################################################
 
 # Version of config file
-version: <<<VERSION>>>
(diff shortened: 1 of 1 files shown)


denyeart:
Close and Re-open to trigger CI.

denyeart:
This PR causes a legitimate test failure in cmd/fabric-ca-server/main_test.go unit test TestErrors.

PR #300, opened 2022-06-17
Bump Alpine to 3.16

denyeart:
Bump Alpine to 3.16.

Signed-off-by: David Enyeart <enyeart@us.ibm.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,7 @@
 #   - vendor - vendors third-party packages
 
 PROJECT_NAME = fabric-ca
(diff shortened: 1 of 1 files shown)


jkneubuh (approved):
This is OK - 👍 

I do wonder if at some point we should move towards standardizing some of the base images.     Why are some on ubuntu, some on alpine, some on RH UBI, etc. etc. etc....?  maybe the introduction of a fabric-base image could help streamline some of the image builds into a common basis.

PR #306, opened 2022-06-27
Correct handling of CA VerifyOptions

mbwhite:
The verify options strucutre is global property - but the verify struture in the case
of re-enroll is set back by 30 seconds.

If this is the global structure - then that is not good.

Duplicating the structure.  Many ways to solve this problem; but this the current approach is cleaner.

Signed-off-by: Matthew B White <whitemat@uk.ibm.com>


jkneubuh (approved):
easy, peasy, lemon squeezy.

👍

PR #311, opened 2022-07-18
updated operations service documentation with /healthz and /version endpoint

basilky:
#### Description

I found that the operations service exposes /healthz and /version endpoint similar to peer and orderer. But it is not mentioned in the documentation. So I updated the documentation with the above endpoints. Content is mostly taken from [fabric documentation](https://github.com/hyperledger/fabric/blob/release-2.2/docs/source/operations_service.rst) of the operations service.


jkneubuh (approved):

denyeart (approved):
Thanks and sorry for the delay!

PR #317, opened 2022-09-29
Github Actions

jcastrence:
#### Type of change

- New feature

#### Description

- Replicate Fabric CA Azure Pipelines on GitHub Actions in preparation to migrate from Azure Pipelines

### DIFF
--- a/ci/azure-pipelines.yml
+++ b/ci/azure-pipelines.yml
@@ -35,17 +35,3 @@ jobs:
     displayName: Checkout Fabric CA Code
   - script: make fvt-tests
     displayName: Run FVT Tests
-
-- job: DocBuild
-  pool:
-    vmImage: ubuntu-latest
-  container:
-    image: n42org/tox:3.4.0
-  steps:
-  - checkout: self
-    path: 'go/src/github.com/hyperledger/fabric-ca'
-    displayName: Checkout Fabric CA Code
-  - script: tox -edocs
-    displayName: Build Documentation
-  - publish: 'docs/_build/html'
-    displayName: Publish Documentation


jcastrence:
For those curious / comparison purposes, here's a [successful run](https://github.com/jcastrence/fabric-ca/actions/runs/3152260589) of the new GitHub Actions workflow.

.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,50 @@
...
+  push:
+    branches: ["**"]
+  pull_request:
+    branches: ["**"]
+  workflow_dispatch:
+
+env:
+  GOPATH: /opt/go
+  PATH: /opt/go/bin:/bin:/usr/bin:/sbin:/usr/sbin:/usr/local/bin:/usr/local/sbin
+  GO_VER: 1.18.2
```
bestbeforetoday, line 16: It might be worth moving to a later Go version, either:
- If GO_VER is used as a complete Go version outside of this workflow, explicitly setting `1.18.6`; or
- If GO_VER is only used in this workflow, using either `1.18` or `'^1.18.6'`

.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,50 @@
...
+
+jobs:
+  verify-build:
+    name: Verify Build
+    runs-on: ubuntu-20.04
```
bestbeforetoday, line 21: It might be a good time to move up to the latest Ubuntu LTS release, now that it has been the active release for a while.
```suggestion
    runs-on: ubuntu-22.04
```

.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,50 @@
...
+    steps:
+      - run: sudo apt-get clean
```
bestbeforetoday, line 23: For several Ubuntu releases, `apt` has been the preferred command for package management with `apt-get` remaining for backwards compatibility. It might be a good time to switch. The commands are the same so it's just a case of replacing `apt-get` with `apt`.

PR #319, opened 2022-10-18
Release Workflow

jcastrence:
#### Type of change

- New feature

#### Description

- Release workflow automates new release creation process similar the current Azure Pipelines release job
- Generates binaries for linux, darwin, and windows targets, pushes Docker images to Hyperledger DockerHub, and creates a new tag and release with artifacts for given version number based off of main branch latest commit
- Workflow must be manually triggered from Actions tab and requires input for the version number and two digit version number similar to Azure Pipelines
- GitHub secrets for DOCKERHUB_PASSWORD and DOCKERHUB_USERNAME must be added to repository matching Hyperledger DockerHub credentials before workflow can be run successfully

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,89 @@
+# Copyright the Hyperledger Fabric contributors. All rights reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+name: Release
+on:
+  workflow_dispatch:
+    inputs:
+      release:
+        description: 'Fabric CA Release'
+        required: true
+        type: string
+      two_digit_release:
+        description: 'Fabric CA Two Digit Release'
+        required: true
+        type: string
+
+env:
+  GO_VER: 1.18
+
+jobs:
+  build-binaries:
+    name: Build Fabric CA Binaries
+    strategy:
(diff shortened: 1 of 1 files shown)


jcastrence:
- [Successful workflow run](https://github.com/jcastrence/fabric-ca/actions/runs/3277065340)
- [Successful release creation](https://github.com/jcastrence/fabric-ca/releases)

mbwhite (approved):
LGTM..

Would it be an idea to make the architecture parameterised - so it's ready for when ARM64 needs to be built?

jcastrence:
> 

Sounds like a good idea to me. So then we would produce 6 total binaries, 3 amd64 and 3 arm64, right? This could be implemented closer to when we'd need to build the arm64 binaries.

jt-nti:
Any chance of arm64 binaries soon? No worries if not, I'll just need to build them myself...
```
Failed to start application: exec: "fabric-ca-server": executable file not found
```

denyeart (approved):
Looks good to me. As a first step let's just replicate the Azure Pipeline and get a release out successfully. Then we can add ARM support.

I'll go ahead and merge.

Issue #321, opened 2022-10-23
Expose Postgres database connection configuration properties including max connection lifetime


kentbull:
Configuring the max lifetime for the Postgres database connection type, and likely all others, like [maxLifetime](https://github.com/brettwooldridge/HikariCP#frequently-used) in HikariCP would allow for regular connection cycling on a configured lifetime duration. I believe the default is "0" in the database/sql library, or infinite lifetimes.

PQ says it "Handles bad connections for database/sql" and [line 727 func TestCloseBadConn](https://github.com/lib/pq/blob/master/conn_test.go#L727) seems to indicate PQ will recover from bad connections, that seem to last for an infinite timeframe, though I have seen in production stale connection pools not recycle stale connections. If we have access to the underlying configuration properties then we can manually configure the connection lifecycle.

The configuration properties I have in mind are as indicated in [this article](https://www.alexedwards.net/blog/configuring-sqldb):

    db.SetMaxOpenConns(25)
    db.SetMaxIdleConns(25)
    db.SetConnMaxLifetime(5*time.Minute)

kentbull:
Here's the relevant code from `database/sql` showing that `If d <= 0, connections are not closed due to a connection's age.`

```
// SetConnMaxLifetime sets the maximum amount of time a connection may be reused.
//
// Expired connections may be closed lazily before reuse.
//
// If d <= 0, connections are not closed due to a connection's age.
func (db *DB) SetConnMaxLifetime(d time.Duration) {
	if d < 0 {
		d = 0
	}
	db.mu.Lock()
	// Wake cleaner up when lifetime is shortened.
	if d > 0 && d < db.maxLifetime && db.cleanerCh != nil {
		select {
		case db.cleanerCh <- struct{}{}:
		default:
		}
	}
	db.maxLifetime = d
	db.startCleanerLocked()
	db.mu.Unlock()
}
```

Looks like all we need to do is expose that API. PR incoming...

PR #324, opened 2022-10-27
Increase stability by exposing DB connection and pool properties 

kentbull:
#### Type of change

#### Description

Exposed database connection properties so we can have automatic connection recycling in production for when connection pools go stale like when Google Cloud SQL does database maintenance about every quarter, sometimes more.

#### Additional details

Similar to the `maxLifetime` configuration property in the [HikariCP](https://github.com/brettwooldridge/HikariCP#essentials) database library, as well as others, providing access to these connection properties adds significant production stability by enabling connection pools to continually refresh themselves. Given that PostgreSQL connections don't garbage collect objects in SQL connections this configurability of connection pools is a requirement for long-term stability of the CA.

#### Related issues

https://github.com/hyperledger/fabric-ca/issues/321

#### Release Note
Exposes operational database connection properties for ease of CA administration and stability.

### DIFF
--- a/lib/caconfig.go
+++ b/lib/caconfig.go
@@ -125,9 +125,13 @@ type CAInfo struct {
 
 // CAConfigDB is the database part of the server's config
 type CAConfigDB struct {
-	Type       string `def:"sqlite3" help:"Type of database; one of: sqlite3, postgres, mysql"`
-	Datasource string `def:"fabric-ca-server.db" help:"Data source which is database specific"`
-	TLS        tls.ClientTLSConfig
+	Type            string `def:"sqlite3" help:"Type of database; one of: sqlite3, postgres, mysql"`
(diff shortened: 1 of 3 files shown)


jkneubuh (changes requested):
Hi @kentbull , thanks for the submission.  This seems like a really nice feature and well worth advancing.  A couple of issues / questions: 

- The unit tests / validation tests are failing with the new interfaces, it looks like they are missing from some test mock APIs.  Please could you make a pass at the tests, as there may be some lingering issues uncovered by the tests.

- The behavior for default should not be to set the values as they are encoded in caconfig.go, but to omit the call to the SetXYZ method if the attribute is not specified in the input configuration.  I.e... leave the behavior "as is," unless the config author has explicitly requested an override of the property.

PR #338, opened 2022-12-02
 Build multi-arch docker images with buildx

jkneubuh:
This PR is configured to publish docker images to ghcr.io/hyperledger/fabric-ca. 

This will allow us to test the mechanics of building, publishing, etc. on a repo other than docker.io prior to the formal 1.5.6 release.

To trigger the release action, draft a new GH Release, applying a semrev tag (e.g. `v1.5.6-beta`) to a target branch. 

Signed-off-by: Josh Kneubuhl <jkneubuh@us.ibm.com>

#### Type of change

- New feature

#### Description

This PR adds support for multi-arch Docker images, generated with the buildx builder.

This PR also modifies the CA binaries to include the `{{ github.ref_name }}` directly when compiling the routine metadata.  This allows for a semantic revision (e.g. `v1.2.3`, `v1.2.3-alpha`, etc.) strings to be specified as valid version identifiers for the binaries. 

#### Additional details

The main challenge with this PR was not the addition of the GHA pipeline to run with buildx, but the dependency on CGO for the platform-specific compilation of sqlite C code.  What unlocked this feature was to break up the "build" into two separate routes: 

- for client binaries, generate a CGO=1 compiled binary, using a GCC cross-compiler to target a specific architecture.
- for Docker images, use Docker `buildx` to emulate an architecture with QEMU, building with sqlite with GCC.

Unfortunately the fabric-ca-fvt images could not be ported over to an arm64 runtime, as there are several dependencies installed to the test image that do not include native support for arm64. When running the fvt tests, the container can be launched on an amd64 machine or under emulation with QEMU.

I tried to clean up the Makefile with this PR, but triggered a never-ending snowball effect that ended up wedging the FVT tests beyond repair.  There may be some lingering dependencies between unit test outputs that have been mounted accidentally and used as inputs to the FVT container volume mounts.

There may still be some lingering rough edges around statically linking the sqlite objects into golang with glibc, which is not technically supported on alpine.

#### Related issues

* [Native Support For 64bit ARM in fabric fabric#2994 (comment)](https://github.com/hyperledger/fabric/issues/2994#issuecomment-1170102505)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,24 +3,16 @@
 # SPDX-License-Identifier: Apache-2.0
 
 name: Release
+
 on:
-  workflow_dispatch:
-    inputs:
-      release:
-        description: 'Fabric CA Release, e.g. 1.5.6'
-        required: true
-        type: string
-      two_digit_release:
-        description: 'Fabric CA Two Digit Release, e.g. 1.5'
-        required: true
-        type: string
-      commit_hash:
-        description: 'Commit hash, e.g. df9c661a192f8cf11376d9d643a0021f1a76c34b'
-        required: true
-        type: string
+  push:
+    tags: [ v* ]
 
 env:
   GO_VER: 1.18.8
+  ALPINE_VER: 3.17
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
 
 permissions:
   contents: read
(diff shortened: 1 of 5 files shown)


.github/workflows/release.yml
```diff
@@ -31,63 +23,104 @@ jobs:
...
       - uses: actions/upload-artifact@v3
         name: Publish Release Artifact
         with:
-          name: hyperledger-fabric-ca-${{ matrix.platform }}-${{ inputs.release }}.tar.gz
-          path: release/${{ matrix.platform }}/hyperledger-fabric-ca-${{ matrix.platform }}-${{ inputs.release }}.tar.gz
+          name: hyperledger-fabric-ca-${{ matrix.platform }}-${{ github.ref_name }}.tar.gz
+          path: release/${{ matrix.platform }}/hyperledger-fabric-ca-${{ matrix.platform }}-${{ github.ref_name }}.tar.gz
+
+  build-and-push-image:
+    runs-on: ubuntu-latest
```
denyeart, line 64: Don't think it matters here, but we usually try to stick to a certain version (currently ubuntu-20.04):
- more consistency
- less confusion
- less chance of behavior changes or breaks when 'latest' changes without our knowledge
- for binaries, less chance of accidentally building binaries on a later version that won't work with earlier ubuntu versions

.github/workflows/release.yml
```diff
@@ -3,24 +3,16 @@
...
-      two_digit_release:
-        description: 'Fabric CA Two Digit Release, e.g. 1.5'
-        required: true
-        type: string
-      commit_hash:
-        description: 'Commit hash, e.g. df9c661a192f8cf11376d9d643a0021f1a76c34b'
-        required: true
-        type: string
+  push:
+    tags: [ v* ]
```
denyeart, line 9: We intentionally picked a workflow_dispatch manual action for a more controlled 'all-in-one' release process, rather than the automagic trigger in the background based on a tag, but let's see what others think.

denyeart (commented):
This is huge progress! Like, one small step for man, one giant leap for mankind huge.

Just a couple nits...

.github/workflows/release.yml
jkneubuh: This actually _is_ supported well by the "all in one" release process. 

With the existing practices, the release process involves manually launching a workflow action / dispatch event.  (There is a step in the workflow that creates the release in GH through a custom action.)

With the "GitHub Way," the order is reversed.  The _human action_ is to create the release (through GUI, etc.), and in response GH will apply a tag to your target branch, which then triggers the release action.

Practically speaking, both models can be supported easily.  The key here in this PR is that the revision metadata is inferred by the TAG triggering the event.  If we want to keep the existing dispatch trigger that's OK - it should just be a single text box entry for the driver to input the semrev value for the release being cranked.

While the order is a little weird, and the "magic tag" seems .... magical ....   it is actually less error prone than manually entering the target for the release, IMO.   I'm OK with either - provided that the input to the process (either way) is a *semrev* string.

Also there are really good mechanics that come from using the implicit branch tags and names.   This took a few hours to grok, but there is some pure gold in the way [docker/metadata-action](https://github.com/docker/metadata-action#typeref) is inferring coordinates from tags and git relative positions.  Good stuff.

BTW the native arm64 performance is insanely fast, compared to QEMU emulation.   Happy to flex the release trigger to however we want it to run.

Hope to see this running on a Raspberry Pi, soon.  (Too bad they are made from _unobtanium!_)

jkneubuh:
The Windows binaries are not working correctly.  Moving to DRAFT status until this can get worked out.

PR #359, opened 2023-05-23
Update operations_guide.rst

thejjw:
FIX: CA Admin identities for org1, org2 must be properly registered as admin

#### Type of change

- Documentation update

#### Description

Following the documentation I got error on [create-and-join-channel](https://hyperledger-fabric-ca.readthedocs.io/en/latest/operations_guide.html#create-and-join-channel) part, getting the following error:
```
Error: got unexpected status: BAD_REQUEST -- error validating channel creation transaction for new channel 'mychannel', could not successfully apply update to template configuration: error authorizing update: error validating DeltaSet: policy for [Group]  /Channel/Application not satisfied: implicit policy evaluation failed - 0 sub-policies were satisfied, but this policy requires 1 of the 'Admins' sub-policies to be satisfied
```
I looked up orderer log to find out that OU being `user` is causing the problem not meeting the policy. I suggest the documentation to use `admin` OU for CA identities of org1 and org2.


denyeart (approved):
Thank you, good catch.

PR #370, opened 2023-08-29
Bump fabric-ca-fvt to bullseye, Bump postgres and mysql

denyeart:
Bump fabric-ca-fvt from debian buster to bullseye,
required since Go 1.21 is only available on the bullseye image.

bullseye requires updating to Postgres 13 and mysql 8.0,
the current LTS releases.

Since the prior databases Postgres 11 and mysql 5.7 are out of support,
update the supported database statements accordingly.

mysql 8.0 changed how to start and stop the db.

Added test failure logging to improve troubleshooting capability.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@ PLATFORM=$(shell go env GOOS)-$(shell go env GOARCH)
 RELEASE_VERSION=$(shell echo $(BASE_VERSION) | sed -e  's/^v\(.*\)/\1/')
 PROJECT_VERSION=${RELEASE_VERSION}
 
-PG_VER=11
+PG_VER=13
 
 PKGNAME = github.com/hyperledger/$(PROJECT_NAME)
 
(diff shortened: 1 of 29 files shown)


denyeart:
@bestbeforetoday Took me a while to figure out all the bullseye and MySql impacts but this one is finally ready.

bestbeforetoday (approved):
I wonder if it might have been worth moving to bookworm (the current stable release) rather than bullseye, since you were going through the pain for moving Debian versions anyway. If you are happy with bullseye then looks OK to me though

denyeart:
> I wonder if it might have been worth moving to bookworm (the current stable release) rather than bullseye, since you were going through the pain for moving Debian versions anyway. If you are happy with bullseye then looks OK to me though

Nothing says we have to stop at bullseye... I just like to move in single jumps so that any unintended impacts can be better understood. I'll go ahead and merge.

PR #384, opened 2023-10-19
Log expiry overrides

denyeart:
If cert expiry during enrollment gets overridden
due to the CA's own expiry, log a message about the override (rather than a debug message).

### DIFF
--- a/lib/serverenroll.go
+++ b/lib/serverenroll.go
@@ -114,24 +114,26 @@ func handleEnroll(ctx *serverRequestContextImpl, id string) (interface{}, error)
 	}
 	req.NotAfter = time.Now().Round(time.Minute).Add(profile.Expiry).UTC()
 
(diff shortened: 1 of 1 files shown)


bestbeforetoday (commented):
I wonder if a warning message might be more appropriate, since it is modifying the requested NotBefore / NotAfter date? Whichever way you prefer is fine; I just wanted to raise the question.

denyeart:
@bestbeforetoday 

> I wonder if a warning message might be more appropriate, since it is modifying the requested NotBefore / NotAfter date? Whichever way you prefer is fine; I just wanted to raise the question.

Sorry for the delay, I only work on CA about once a month :-)

Agreed, changed to Warning. And suppressed the Warning if NotBefore is not set so that the log doesn't fill up with superfluous warnings for every request.

PR #404, opened 2024-02-13
Use fabric-lib-go for bccsp, metrics, flogging

denyeart:
The following packages are being moved from fabric to fabric-lib-go so that they can be shared across fabric and fabric-ca:

- github.com/hyperledger/fabric/bccsp
- github.com/hyperledger/fabric/common/flogging
- github.com/hyperledger/fabric/common/metrics

This commit updates fabric-ca to use the common code in fabric-lib-go.

The fabric-ca dependency on fabric is no longer needed. That's good!

Also ran goimports to clean up imports.


denyeart:
Go approval in fabric repo to push down the dependency to fabric-lib-go, I'll go ahead and merge.

PR #410, opened 2024-03-30
handling `setLogLevel` error of `ConfigInit` if both `debug` flag and `loglevel` are set

bkiran6398:
#### Type of change
- Bug fix

#### Description
Either `debug` flag or a valid `loglevel` string is expected to be set to determine the log level of application.
When both are set, `setLogLevel` method returns an error which is handled in `ConfigInit`.


denyeart:
@mergifyio rebase

denyeart:
See the checks for a legitimate unit test failure caused by this change.

PR #424, opened 2024-07-01
Update cadeploy.md

atomic-kanta-sasaki:
fix:  command registers the organization CA bootstrap identity rcaadmin with password rcaadminpw with the TLS CA.

#### Type of change

- Documentation update

#### Description

I found and fixed a command that did not work when I was trying the section `Register and enroll the organization CA bootstrap identity with the TLS CA`.


denyeart (commented):
Actually for the 'register' command, the admin client is authenticated using the credentials in the --mspdir flag. There is no need to pass it separately in the -u flag.

For the 'enroll' command, you pass the enroller credentials in the -u flag.

What you've done may work, but to be consistent with the rest of this doc I think we should leave the authentication information out of the register -u flag.

You can look at all the other references of --mspdir in this doc to better understand the documented and intended use.

Perhaps you had --mspdir incorrect which is why it didn't work for you the first time?

atomic-kanta-sasaki:
@denyeart 
I ran it again and found that to be true.
Please confirm that I have corrected the problem.

PR #430, opened 2024-08-08
Update Go version in README

denyeart:
Update Go version in README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ The remainder of this guide is intended for developers contributing to Fabric CA
 
 ## Prerequisites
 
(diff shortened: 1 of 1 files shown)


bestbeforetoday (approved):
You might want to rename the commit as "Rename Go version in README" (rather than Docker version) on merge since that's what changed. Otherwise looks good

denyeart:
> You might want to rename the commit as "Rename Go version in README" (rather than Docker version) on merge since that's what changed. Otherwise looks good

Thanks... I was going to update both... fixed to indicate just Go now!

PR #433, opened 2024-09-20
Release commit for v1.5.13

denyeart:
Add release notes for v1.5.13.


bestbeforetoday (approved):
LGTM.

Looking at the one known vulnerability in these release notes, it seems like it might be time to mitigate this by changing the default value for `FABRIC_CA_SERVER_COMPATIBILITY_MODE_V1_3` to `false`. The text says we would do this once we moved from Fabric v1 to v2.

bestbeforetoday:
I raised #434 for the default compatibility mode change.

PR #444, opened 2025-02-07
Add config debug

denyeart:
Changing V1_3 compatibility mode in https://github.com/hyperledger/fabric-ca/pull/435  has caused some user confusion, some additional debug logging will help:

- Info log FABRIC_CA_SERVER_COMPATIBILITY_MODE_V1_3 setting
- Debug log when falling back to COMPATIBILITY_MODE_V1_3 authentication header validation

Also take this opportunity to log the entire config upon startup to be consistent with core Fabric:

- Info log CA config upon startup

### DIFF
--- a/cmd/fabric-ca-server/config.go
+++ b/cmd/fabric-ca-server/config.go
@@ -9,6 +9,7 @@ package main
 import (
 	"os"
 	"path/filepath"
+	"slices"
 	"strings"
 
 	"github.com/hyperledger/fabric-ca/lib/common/idemix"
@@ -19,6 +20,7 @@ import (
 	"github.com/hyperledger/fabric-ca/lib/metadata"
 	"github.com/hyperledger/fabric-ca/util"
 	"github.com/pkg/errors"
+	"gopkg.in/yaml.v2"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


bestbeforetoday (approved):

bestbeforetoday (approved):

cmd/fabric-ca-server/config.go
```diff
@@ -578,6 +579,20 @@ func (s *ServerCmd) configInit() (err error) {
 		return err
 	}
 
+	// Info logging for CA config, includes yaml file settings and environment variable overrides
+	allSettings := s.myViper.AllSettings()
+	settingsYaml, err := yaml.Marshal(allSettings)
+	if err == nil {
+		log.Infof("CA config with combined yaml file and environment variable overrides:\n%s", settingsYaml)
```
bestbeforetoday, line 586: It seems the configuration can contain passwords, which is failing an FVT test that the log doesn't leak passwords

cmd/fabric-ca-server/config.go
```diff
@@ -578,6 +579,20 @@ func (s *ServerCmd) configInit() (err error) {
...
+	}
+
+	// Debug logging for CA environment variables
+	log.Debugf("Environment variables:")
+	envVars := os.Environ()
+	for _, envVar := range envVars {
+		log.Debug(envVar)
+	}
```
bestbeforetoday, line 594: This can also leak passwords in environment variables.

denyeart: ok, I've removed the env variables.

cmd/fabric-ca-server/config.go
denyeart: Ok, I've filtered the config of sensitive configurations now.

bestbeforetoday (approved):

PR #446, opened 2025-02-08
Separate runner for amd64 and arm64

denyeart:
Previously arm64 images were created on a standard amd64 runner using qemu. qemu emulation is very slow and there have been recent issues. Native image creation is faster and more reliable.

This change creates separate runners for amd64 (ubuntu-22.04) and arm64 (ubuntu-22.04-arm). Since ubuntu-20.04-arm runner is not available, binaries and images will be built with ubuntu-22.04.

A new job is then needed to combine the image digests into a multi-architecture manifest that gets pushed to Docker Hub and ghcr.

The approach is copied from core fabric repository.

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -21,7 +21,7 @@ permissions:
 jobs:
   verify-build:
     name: Verify Build
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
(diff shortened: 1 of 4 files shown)


denyeart:
Since release approach is already used in core fabric, and no code changes made, I'll go ahead and merge and release v1.5.15.

PR #458, opened 2025-07-15
Decouple Cred Generation from file system depnedencies and side effects

jmbarzee:
#### Type of change

#### Description
Just enables the client packages to be used as libraries without the file system side effects. This is critical for workflow systems that leverage ephemeral workflows to manipulate or rotate credentials.

#### Additional details
Changes are fully backwards compatible, and are covered entirely by existing tests.

#### Related issues
N/A

#### Release Note
lib/client now supports certificate generation without file system side effects.

Checklist (Pending Items):
- `Signed-off-by` added to commits (required for DCO check to pass)


bestbeforetoday (commented):
Thank you for the contribution. Your changes look good to me, and the code structure is nice and clean.

My primary concern is that this module is not really intended to be consumed as a library. It is the implementation of the Fabric CA (binary) application and accompanying CLI commands. This change makes it more usable as a library, and I am not sure this is something we should be supporting. What do you think @denyeart?

PR #489, opened 2026-04-05
[fabric-ca-client] Added CLI flags to write enrollment output files at custom paths

pasquale95:
#### Type of change

- New feature

#### Description

`fabric-ca-client enroll` writes output files to fixed locations within the MSP directory structure: the private key is stored by the BCCSP keystore under a hash-derived filename (`<ski_hex>_sk`), the enrollment certificate at `signcerts/cert.pem`, and the CA certificate at `[tls]cacerts/<server-url-derived>.pem`. These locations are not configurable and not in line with what `cryptogen` usually generates, making harder the switch from such development tool to `fabric-ca`.

Since we use Fabric-CA within our Fabric-X deployment scripts, we have currently _patched_ this behaviour introducing some extra-steps summarized in the [`cryptogenize.yaml`](https://github.com/LF-Decentralized-Trust-labs/fabric-x-ansible-collection/blob/main/roles/fabric_ca/tasks/client/cryptogenize.yaml) task. However this task introduces unnecessary overhead that we could simply avoid by enhancing the current `fabric-ca-client` binary.

This change adds three optional flags to `ClientConfig`, available to the `enroll` and `reenroll` commands:

- `--myskfile <path>` — path relative to the MSP directory where the private key is written
  (e.g. `server.key` produces `<mspdir>/server.key`)
- `--mycertfile <path>` — path relative to the MSP directory where the enrollment certificate
  is written (e.g. `server.crt` produces `<mspdir>/server.crt`)
- `--mycacertfile <path>` — path relative to the MSP directory where the CA certificate is
  written (e.g. `ca.crt` produces `<mspdir>/ca.crt`)

When a flag is set, the file is written at the specified path instead of the default MSP subdirectory. When a flag is not set, existing behavior is fully preserved.

The flags are registered automatically via `util.RegisterFlags` using the existing struct-tag reflection mechanism, consistent with all other `ClientConfig` fields. They are also bindable via environment variables (`FABRIC_CA_CLIENT_MYSKFILE`, etc.) and the configuration file.

#### Additional details

We have set the flags as `my...` to stick with the convention that we have seen with the flag `--myhost` (which sets the hostname in the CSR). It is used to distinguish output/identity-related flags from connection and TLS flags
(`--tls.client.certfile`, `--tls.client.keyfile`, `--tls.certfiles`), which operate on different files and would be ambiguous with shorter names like `--certfile` or `--keyfile`.

#### Note

After the merge of this PR it would be great if you could release a new version (e.g. `1.5.19`) since we use the Docker images in our deployment scripts and to leverage this new functionality we would need `fabric-ca-client:1.5.19` image.

Thank you!

### DIFF
--- a/cmd/fabric-ca-client/command/getcainfo.go
+++ b/cmd/fabric-ca-client/command/getcainfo.go
@@ -156,10 +156,16 @@ func storeCAChain(config *lib.ClientConfig, si *lib.GetCAInfoResponse) error {
 		}
 	}
 
-	// Store the root certificates in the "cacerts" msp folder
+	// Store the root certificates in the "cacerts" msp folder, or at the custom path if set.
 	certBytes := bytes.Join(rootBlks, []byte(""))
 	if len(certBytes) > 0 {
-		if config.Enrollment.Profile == "tls" {
+		if config.MyCACertFile != "" {
+			dst := path.Join(mspDir, config.MyCACertFile)
+			err := storeToFile("CA certificate", path.Dir(dst), path.Base(dst), certBytes)
+			if err != nil {
+				return err
+			}
+		} else if config.Enrollment.Profile == "tls" {
 			err := storeToFile("TLS root CA certificate", tlsRootCACertsDir, tlsfname, certBytes)
 			if err != nil {
 				return err
--- a/cmd/fabric-ca-client/command/getcainfo.go
+++ b/cmd/fabric-ca-client/command/getcainfo.go
@@ -160,8 +160,15 @@ func storeCAChain(config *lib.ClientConfig, si *lib.GetCAInfoResponse) error {
 	certBytes := bytes.Join(rootBlks, []byte(""))
 	if len(certBytes) > 0 {
 		if config.MyCACertFile != "" {
-			dst := path.Join(mspDir, config.MyCACertFile)
-			err := storeToFile("CA certificate", path.Dir(dst), path.Base(dst), certBytes)
+			dst, err := util.MakeFileAbsWithinDir(config.MyCACertFile, mspDir)
+			if err != nil {
+				return err
+			}
+			desc := "root CA certificate"
+			if config.Enrollment.Profile == "tls" {
+				desc = "TLS root CA certificate"
+			}
+			err = storeToFile(desc, filepath.Dir(dst), filepath.Base(dst), certBytes)
 			if err != nil {
 				return err
 			}
--- a/lib/client.go
+++ b/lib/client.go
@@ -111,7 +111,15 @@ func (c *Client) Init() error {
 		if err != nil {
 			return errors.Wrap(err, "Failed to create signcerts directory")
 		}
-		c.certFile = path.Join(certDir, "cert.pem")
+		if cfg.MyCertFile != "" {
+			certFile := path.Join(mspDir, cfg.MyCertFile)
+			if err = os.MkdirAll(path.Dir(certFile), 0o755); err != nil {
+				return errors.Wrap(err, "Failed to create directory for cert file")
+			}
+			c.certFile = certFile
+		} else {
+			c.certFile = path.Join(certDir, "cert.pem")
+		}
 
 		// CA certs directory
 		c.caCertsDir = path.Join(mspDir, "cacerts")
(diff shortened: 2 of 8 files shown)


pasquale95:
@denyeart @bestbeforetoday could you please have a look to this PR and let me know if there's anything else to do in order to merge it?

C0rWin (dismissed):

pasquale95:
Hi @C0rWin, thanks for the approval! @bestbeforetoday @denyeart could you please merge this PR into `main` and release a new tag version?

cmd/fabric-ca-client/command/getcainfo.go
```diff
@@ -156,10 +156,16 @@ func storeCAChain(config *lib.ClientConfig, si *lib.GetCAInfoResponse) error {
 		}
 	}
 
-	// Store the root certificates in the "cacerts" msp folder
+	// Store the root certificates in the "cacerts" msp folder, or at the custom path if set.
 	certBytes := bytes.Join(rootBlks, []byte(""))
 	if len(certBytes) > 0 {
-		if config.Enrollment.Profile == "tls" {
+		if config.MyCACertFile != "" {
+			dst := path.Join(mspDir, config.MyCACertFile)
```
bestbeforetoday: Are we happy with any filesystem path relative to the `mspDir` or should the destination be strictly within the `mspDir`? If the latter, there needs to be a some different handling to avoid path escapes.

cmd/fabric-ca-client/command/getcainfo.go
```diff
@@ -156,10 +156,16 @@ func storeCAChain(config *lib.ClientConfig, si *lib.GetCAInfoResponse) error {
...
+			err := storeToFile("CA certificate", path.Dir(dst), path.Base(dst), certBytes)
```
bestbeforetoday: This should probably say `"root CA certificate"` for consistency with the naming below. Or maybe also include "tls" depending the enrollment profile?

cmd/fabric-ca-client/command/getcainfo.go
bestbeforetoday, line 188: This change allows the root certificate location to be changed but not the intermediate certificates. What does the workflow look like for end users wanting a custom cert location but where there are intermediate certs?

lib/client.go
```diff
@@ -111,7 +111,15 @@ func (c *Client) Init() error {
 		if err != nil {
 			return errors.Wrap(err, "Failed to create signcerts directory")
 		}
-		c.certFile = path.Join(certDir, "cert.pem")
+		if cfg.MyCertFile != "" {
+			certFile := path.Join(mspDir, cfg.MyCertFile)
```
bestbeforetoday: Should we consider path escapes here?

lib/client.go
```diff
@@ -321,12 +329,36 @@ func (c *Client) net2LocalCAInfo(net *api.CAInfoResponseNet, local *GetCAInfoRes
 	return nil
 }
 
+// storeCustomKeyFile moves the BCCSP-managed private key to a custom path relative to the MSP
+// directory, if MySkFile is configured.
+func (c *Client) storeCustomKeyFile(key bccsp.Key) error {
+	if c.Config.MySkFile == "" {
+		return nil
+	}
+	src := path.Join(c.Config.MSPDir, "keystore", hex.EncodeToString(key.SKI())+"_sk")
```
bestbeforetoday: This looks pretty fragile.
1. It relies on knowledge of the keystore location, which is scattered around the codebase all using the hard-coded "keystore" directory name. While it is already messy, I don't like the idea of adding more mess. Perhaps this could be factored out to some common code.
2. It also relies on knowledge of some implementation details from fabric-lib-go, which could change independently of this code and break things.

lib/client.go
```diff
@@ -321,12 +329,36 @@ func (c *Client) net2LocalCAInfo(net *api.CAInfoResponseNet, local *GetCAInfoRes
...
+	return os.Remove(src)
+}
+
 func (c *Client) handleX509Enroll(req *api.EnrollmentRequest) (*EnrollmentResponse, error) {
 	// Generate the CSR
 	csrPEM, key, err := c.GenCSR(req.CSR, req.Name)
 	if err != nil {
 		return nil, errors.WithMessage(err, "Failure generating CSR")
 	}
+	if err := c.storeCustomKeyFile(key); err != nil {
```
bestbeforetoday, line 381: I wonder if the cleanest approach might be to modify fabric-lib-go so that private keys actually have a usable `Bytes` implementation that can be used here to get the private key bytes directly from the `key` object instead of having to read from the underlying keystore.

bestbeforetoday (commented):
I am very uncomfortable that this change has zero accompanying tests. It does not give much confidence that the current change works, and certainly does not help to avoid the new functionality getting accidentally broken by future changes. Manual testing is not viable.

cmd/fabric-ca-client/command/getcainfo.go
pasquale95: Strictly under `mspDir` is indeed what we need. Fixed.

cmd/fabric-ca-client/command/getcainfo.go
pasquale95: Added better descriptive message for both MSP and TLS enrollments.

cmd/fabric-ca-client/command/getcainfo.go
pasquale95: Intermediate Certs are currently out of scope for this PR. As indicated in the PR description, the goal of this PR is to provide users with the possibility to use `fabric-ca-client` to produce `cryptogen-alike` crypto material. Since `cryptogen` doesn't produce intermediate certs, I've not touched that part at all.

lib/client.go
pasquale95: Done.

lib/client.go
pasquale95: Modifying `fabric-lib-go` is out-of-scope for this PR. I've however refactored the code to have a central function used to retrieve the `mspKeystoreDir` as per your suggestion.

lib/client.go
pasquale95: As stated above, modifying `fabric-lib-go` is currently out-of-scope, but nothing prevents to later modify it and improve this code. However, given the current lib limitations, I believe this is the only approach that we can use.

pasquale95:
Hi @bestbeforetoday, thanks for the comments.
I've gone through all of them and:
- added escape path validation for safety;
- added 2 new tests to validate the changes: `TestEnrollCustomOutputFiles` and `TestEnrollCustomOutputFilesRejectPathEscapes`;
- minor refactoring around the keystorePath to have it centralized as per your suggestion.

Regarding your suggestion to change `fabric-lib-go`, I'm not against it (I was at the beginning checking whether i could retrieve the key bytes and write them directly, cause I also believe is a cleaner approach). However I cannot assess why the private key BCCSP `KeyGen` don't return the `Bytes()` (maybe a security concern?). Hence, I believe that the current approach is the safest one in terms of full backward-compatibilty.

Nothing prevents in the future to adjust these functions to use the Bytes() if they will ever be exposed.

bestbeforetoday (approved):
Many thank for the updates, including the tests. It looks good to go for me now.

PR #506, opened 2026-06-05
Fix client command detection when flags precede subcommand

### LINKED ISSUE #465: fabric-ca-client: validation issues when flags are positioned before the subcommand
Some validation logic in `fabric-ca-client` is being executed based on the value of the first CLI argument, which is erroneously assumed to always be a subcommand, when it can be a flag instead. This can cause mystifying problems in some situations, especially given that the ability to place flags before the subcommand (as is conventionally expected of global flags) otherwise works as expected and is quite convenient for writing shell aliases.

### Example

When running the `enroll` subcommand with an (unsurprisingly) empty MSP directory, the following error will be produced:

```
Error: Enrollment information does not exist. Please execute enroll command first. Example: fabric-ca-client enroll -u http://user:userpw@serverAddr:serverPort
```

This happens whenever **any** flag is positioned **before** the subcommand `enroll`, such as: 

```fabric-ca-client --loglevel warning enroll --url (...)```

In this situation, the command will only succeed in two situations.

1. If `enroll` is the very first argument

```fabric-ca-client enroll --loglevel warning --url (...)```

2. If the target MSP directory is already populated, so that the previously failing validation passes.

### Relevant code

https://github.com/hyperledger/fabric-ca/blob/00f89fdf1dc8dbd8600efbc191e94b7c145ad80b/cmd/fabric-ca-client/command/root.go#L28-L32
https://github.com/hyperledger/fabric-ca/blob/00f89fdf1dc8dbd8600efbc191e94b7c145ad80b/cmd/fabric-ca-client/command/clientcmd.go#L116-L121
https://github.com/hyperledger/fabric-ca/blob/00f89fdf1dc8dbd8600efbc191e94b7c145ad80b/cmd/fabric-ca-client/command/clientcmd.go#L243-L254

### PR DESCRIPTION
**Pull request by antcybersec**
## Summary

- Resolve the active subcommand with cobra's `Find` instead of assuming `args[1]` is always the command name
- Fixes incorrect enrollment validation when global flags (e.g. `--loglevel warning`) appear before `enroll`, `gencsr`, `getcacert`, or `getcainfo`
- Add unit tests for command resolution and the reported enroll failure mode

## Problem

`fabric-ca-client --loglevel warning enroll -u ...` was treated as if the subcommand was `warning`, so `ConfigInit` incorrectly required prior enrollment and failed with:

> Enrollment information does not exist. Please execute enroll command first.

## Test plan

- [x] `go test ./cmd/fabric-ca-client/command/ -run 'TestResolveCommandName|TestEnrollWithGlobalFlagsBeforeSubcommand'`
- [x] `go test ./cmd/fabric-ca-client/command/`

Fixes #465

### DIFF
--- a/cmd/fabric-ca-client/command/root_test.go
+++ b/cmd/fabric-ca-client/command/root_test.go
@@ -1,0 +1,77 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package command
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestResolveCommandName(t *testing.T) {
+	ccmd := NewCommand("")
+
+	testCases := []struct {
+		name     string
+		args     []string
+		expected string
+	}{
+		{
+			name:     "subcommand first",
+			args:     []string{"enroll", "-u", "http://admin:adminpw@localhost:7054"},
+			expected: "enroll",
+		},
+		{
+			name:     "global flags before subcommand",
+			args:     []string{"--loglevel", "warning", "enroll", "-u", "http://admin:adminpw@localhost:7054"},
+			expected: "enroll",
+		},
+		{
+			name:     "global flag with equals before subcommand",
+			args:     []string{"--loglevel=warning", "gencsr", "--csr.cn", "identity"},
+			expected: "gencsr",
+		},
+		{
+			name:     "nested subcommand",
+			args:     []string{"--loglevel", "warning", "identity", "list"},
+			expected: "identity",
+		},
+		{
+			name:     "getcacert alias",
+			args:     []string{"--home", "/tmp/home", "getcacert", "-u", "http://localhost:7054"},
+			expected: "getcainfo",
+		},
+		{
+			name:     "no subcommand",
+			args:     []string{"--loglevel", "warning"},
+			expected: "",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			assert.Equal(t, tc.expected, resolveCommandName(ccmd.rootCmd, tc.args))
+		})
+	}
+}
+
+func TestEnrollWithGlobalFlagsBeforeSubcommand(t *testing.T) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: cmd/fabric-ca-client/command/root_test.go [hunk 1]
```diff
@@ -0,0 +1,74 @@
...
+func TestEnrollWithGlobalFlagsBeforeSubcommand(t *testing.T) {
+	defYaml := util.GetDefaultConfigFile("fabric-ca-client")
+	os.Remove(defYaml)
+	defer os.Remove(defYaml)
+
+	err := RunMain([]string{cmdName, "--loglevel", "warning", "enroll", "-d"})
+	if assert.Error(t, err) {
+		assert.NotContains(t, err.Error(), "Enrollment information does not exist",
+			"enroll with flags before subcommand should not require prior enrollment")
+	}
```
bestbeforetoday: I think this configuration should be guaranteed to fail so it should require an error, not just assert if an error occurs. The NotContians assertion on the error message also seems error-prone since the error messages could be changed in the future, which would make this test pass regardless of actual behaviour.

I think this really needs to change to a positive test:


```suggestion
	adminHome := filepath.Join(tdDir, "enrolladminhome")

	// Remove admin home directory if it exists
	err := os.RemoveAll(adminHome)
	require.NoErrorf(t, err, "Failed to remove directory %s: %s", adminHome, err)

	// Remove admin home directory that this test is going to create before
	// exiting the test case
	defer os.RemoveAll(adminHome)

	srv := setupEnrollTest(t)

	// Cleanup before exiting the test case
	defer stopAndCleanupServer(t, srv)

	err = RunMain([]string{cmdName, "--loglevel", "warning", "enroll", "-d", "-u", enrollURL, "-H", adminHome})
	require.NoError(t, err, "enroll should succeed with global flags before subcommand")
```

bestbeforetoday (changes requested):
Thank you very much for this contribution. Just a concern with one of the unit tests (see inline comments); otherwise looks good to me.

cmd/fabric-ca-client/command/root_test.go
antcybersec: Good point — updated to a positive test that enrolls against a running CA server with global flags before the subcommand. Thanks for the suggestion.

bestbeforetoday (approved):

PR #509, opened 2026-07-02
Add support for `s390x` binary and image build in Github Actions

pasquale95:
#### Type of change

#### Description

This PR aims to add the `s390x` platform among the supported platforms for the Docker multiplatform image and also build the `s390x` binary using GitHub Actions.

With this PR, the pipeline will now generate:

- binaries: `linux/amd64`, `linux/arm64`, `linux/s390x`, `darwin/arm64`, `windows/amd64`;
- image: `linux/amd64`, `linux/arm64`, `linux/s390x`.

#### Additional details

For the images I've moved the workflow from native-VM to QEMU-emulation since GitHub Actions doesn't provide a native `s390x` runner. It follows the same approach that we use in [`fabric-x-orderer`](https://github.com/hyperledger/fabric-x-orderer/blob/main/.github/workflows/docker-release.yml) for the multi-platform release.

The artifacts have been correctly generated using a [test tag in my fork](https://github.com/pasquale95/fabric-ca/pkgs/container/fabric-ca).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,6 +26,8 @@ jobs:
             platform: linux-amd64
           - image: ubuntu-22.04
             platform: linux-arm64
+          - image: ubuntu-22.04
+            platform: linux-s390x
           - image: macos-latest
             platform: darwin-arm64
           - image: macos-15-intel
@@ -48,6 +50,7 @@ jobs:
           sudo apt-get update
           sudo apt-get -y install gcc-aarch64-linux-gnu
           sudo apt-get -y install gcc-x86-64-linux-gnu
+          sudo apt-get -y install gcc-s390x-linux-gnu
 
       - run: make dist/${{ matrix.platform }}
         name: Compile Binary and Create Tarball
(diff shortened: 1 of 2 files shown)


pasquale95:
@denyeart @bestbeforetoday could you quickly have a look to this PR and if you like it, approve it?

adecaro (approved):
LGTM

.github/workflows/release.yml
```diff
@@ -74,10 +78,6 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        runner:
-          - ubuntu-22.04 # creates linux-amd64 images
-          - ubuntu-22.04-arm # creates linux-arm64 images
```
bestbeforetoday, line 79: We specifically used amd64 and arm runners rather than qemu to build natively for those target architectures because of a bug in a previous version of qemu that failed builds

bestbeforetoday (commented):
This change has changed the build behaviour. Provided the qemu bugs don't return and the publishing actually works, that should be fine.

Did you test publishing to the GitHub Docker registry by creating a release in your fork?

PR #510, opened 2026-07-09
Depend on the monolithic idemix v0.2.0 module

EvanYan1024:
#### Type of change

#### Description

fabric-ca currently depends on the *split* idemix layout — `github.com/IBM/idemix v0.0.2-...` plus the standalone `github.com/IBM/idemix/bccsp/types` and `github.com/IBM/idemix/bccsp/schemes/weak-bb` modules.

Both the split sub-modules and the monolithic `github.com/IBM/idemix` module export the `github.com/IBM/idemix/bccsp/types` import path. As a result, any module graph that pulls fabric-ca alongside a dependency that requires the monolithic idemix fails to build with an ambiguous import, e.g.:

```
ambiguous import: found package github.com/IBM/idemix/bccsp/types in multiple modules:
	github.com/IBM/idemix v0.1.1 (.../idemix@v0.1.1/bccsp/types)
	github.com/IBM/idemix/bccsp/types v0.0.0-20240913182345-72941a5f41cd (.../bccsp/types@...)
```

This blocks using the fabric-ca client library in the same build as recent `fabric-token-sdk` / LFDT Panurus releases, which depend on the monolithic `github.com/IBM/idemix v0.1.1`.

This change depends on the monolithic `github.com/IBM/idemix v0.1.1` and drops the split `bccsp/types` and `bccsp/schemes/weak-bb` requirements; `github.com/IBM/mathlib` is bumped to `v0.2.0` to match. v0.1.1 folds the sub-packages back into the single module, so the `github.com/IBM/idemix/bccsp/...` import paths still resolve.

#### Additional details

No source changes — only `go.mod` / `go.sum` and the regenerated `vendor/` tree. fabric-ca compiles unchanged against idemix v0.1.1; verified locally with `go build ./...`.

#### Related issues

N/A


bestbeforetoday:
fabric-ca is an application, distributed as a binary. It is not intended to be consumed as a Go library.

Rationalising the imports is fine but it must remain exactly compatible with [Fabric](https://github.com/hyperledger/fabric).

go.mod
```diff
@@ -1,13 +1,12 @@
 module github.com/hyperledger/fabric-ca
 
-go 1.25.0
+go 1.26.3
```
bestbeforetoday, line 3: I guess this is required by the newer idemix version. We had deliberately left it back at the lowest currently supported Go version to avoid strict Go version requirements for people already using fabric-ca as a Go library but, since fabric-ca is not intended to be used that way, this change seems OK.

go.mod
```diff
@@ -1,13 +1,12 @@
...
 
 toolchain go1.26.4
 
 require (
-	github.com/IBM/idemix v0.0.2-0.20240913182345-72941a5f41cd
-	github.com/IBM/idemix/bccsp/types v0.0.0-20240913182345-72941a5f41cd
-	github.com/IBM/mathlib v0.0.3-0.20250709075152-a138079496c3
+	github.com/IBM/idemix v0.1.1
```
bestbeforetoday: There looks to be a v0.2.0 release now.

go.mod
```diff
@@ -1,13 +1,12 @@
...
+	github.com/IBM/mathlib v0.2.0
```
bestbeforetoday: There is now a v0.3.0 release, which is a dependency of github.com/IBM/idemix@v0.2.0

EvanYan1024:
> fabric-ca is an application, distributed as a binary. It is not intended to be consumed as a Go library.
> 
> Rationalising the imports is fine but it must remain exactly compatible with [Fabric](https://github.com/hyperledger/fabric).

Thanks. For context, we don't run the fabric-ca binary — we embed the client library (lib.Client + api) for programmatic per-institution enroll/register, which the CLI can't cover.

Fully agree it must stay in lockstep with Fabric. The trigger here: fabric-token-sdk (LFDT Panurus) and fabric-smart-client moved to the monolithic github.com/IBM/idemix, which can't coexist with fabric-ca's split bccsp/types in one build (ambiguous import). Single-module is also the current idemix direction — v0.1.0/v0.1.1 and today's v0.2.0 are all single-module, while Fabric is still on the 2024 split commit. So fabric-ca will need this realignment whenever Fabric refreshes idemix; this just does it early.

go.mod
EvanYan1024: Right, idemix v0.1.1 pulls in go 1.26 so tidy bumped it.

go.mod
EvanYan1024: Good catch, will move to idemix v0.2.0.

go.mod
EvanYan1024: Will bump mathlib to v0.3.0 to match idemix v0.2.0.

bestbeforetoday (approved):

PR #512, opened 2026-07-24
WIP: Update fabric-lib-go and viper dependencies

bestbeforetoday:
Changes were required to:

- Avoid conflicting usage of github.com/go-viper/mapstructure/v2 and github.com/mitchellh/mapstructure.
- Adopt Viper API changes.
- Prevent fvt tests from creating invalid YAML with duplicate keys, which were tolerated by older YAML parser versions but now cause a failure.


bestbeforetoday:
Superseded by #521

